PR #14, opened 2015-11-01
LUI-35

jdegraft:
mvc approach to moving index pages


jdegraft:
@dkayiwa 
I don't know which file you mean.
with this approach we don't need AdminPageController.java because the index page is now mapped in the config file. AdminPageController was added by mean in an earlier ticket to map admin/index.htm so it can now be removed.

dkayiwa:
I mean jsp pages that are not index pages.

jdegraft:
@dkayiwa 
I was not sure what to do.
The ticket said to move dictionary index.
The folder containing dictionary index did not exist in legacyui.
So I thought I might as well move them as well. Technically only the index page should be moved as part of this ticket.

dkayiwa:
I would propose that we separate the two. That way i can easily merge the mere moving of the existing jsp pages, even if the index thing is not yet accepted to be merged.

jdegraft:
@dkayiwa 

ok, I will remove the additional jsps which are not index pages. This will only leave one index page I moved dictionary/index because the other index.jsp pages were already in legacyui.

jdegraft:
@dkayiwa 
updated pull request.
my memory failed me, actually as part of the ticket I had to moved several index pages, so those are still part of this pull request.
I have removed the additional jsps which are not index pages in the dictionary folder which the ticket did not require that I move.

jdegraft:
@dkayiwa 
I did not mean to remove any beans.
I used the git process steps you gave me a few weeks ago to squash commits and this generated the pull request. (Maybe it is part of another commit and is a reflection of the rebase.)

The only thing I have added is the xsd definition for mvc
and mvc:controller tags mapping url - views

I can redo the request if this is the approach we want to go with. It was just a research activity I wanted to share. (The other pull request with classes for each url - view mapping is still available as a contrast.)

dkayiwa:
So can we add it back?

jdegraft:
yes

jdegraft:
@dkayiwa. ok, I will put back the bean. Is that the comment you are talking about?

dkayiwa:
Yes!

On Mon, Nov 2, 2015 at 5:59 PM, jdegraft <notifications@github.com> wrote:

> @dkayiwa <https://github.com/dkayiwa>. ok, I will put back the bean. Is
> that the comment you are talking about?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/openmrs/openmrs-module-legacyui/pull/14#issuecomment-153043333>
> .
>



-- 
If we keep uppermost in our minds the unkind and unjust acts of others, we
shall find it impossible to love them as Christ has loved us; but if our
thoughts dwell upon the wondrous love and pity of Christ for us, the same
spirit will flow out to others.

jdegraft:
@dkayiwa I have put the bean back. There some changes which Wyclif added which I have put back as well.

omod/src/main/resources/webModuleApplicationContext.xml
```diff
@@ -129,11 +139,39 @@
...
-				
+
  			</props>
 		</property>
 	</bean>	
 
+ 	<!-- index pages - begin -->  
+	
+	<!-- admin -->
+	<mvc:view-controller view-name="module/legacyui/admin/index" path="admin/index.htm"/>
```
wluyima: I doubt if this will work for localhost:8080/opemmrs/admin

jdegraft: Good point. I was not sure whether I was supposed to map /admin as well /admin/index. The way it is setup only /admin/index is mapped. Should I map /admin as well?

wluyima: The ticket actually was meant to get the URLs like the one i added in my previous comment to work, if you look at the ticket description, i listed the URLS that need to be fixed

jdegraft: @wluyima, we have a problem.
I have spent several days on this ticket and this is what seems to be happening.
1st we have to distinguish between a url and a view.
From what I have found a url is processed by a controller to return a view.
The way openmrs (spring) is setup, "/admin" is considered as view, so there is no way to use a controller to map it to a different view eg under "module/legacyui".
To achieve that, we have to use view resolution, which is a bit tricky as it is easy to break the rest of the application.
(openmrs already has a complicated view resolution setup which basically resolves  "/admin" to "WEB-INF/view/admin")
In other words, the parameterizable controllers are not getting a chance to process "/admin".
I am sure this would be the same effect if I wrote a class using "Controller" and "RequestMapping" as that is the code which the paramterizable controller generates anyway.
Do you agree? If so what approach would you recommend.

wluyima: I'm actually already aware of this, the ideal solution was to map these urls to the Dispatcher servlet from the module, but even the servlet API provides no way to remove a servlet in case the module is stopped/removed. In interest of time, under dev notes on the ticket the only quick solution to find usage of these URLs and change them e.g /openmrs/admin becomes /openmrs/admin/index.jsp

jdegraft: One final question.
Do I restrict these changes to index pages or in general?
For instance if there is a reference to /admin/patientForm.jsp do I change it /module/legacyui/patientForm.jsp or leave that to be done in another ticket?

dkayiwa: Do as the ticket says. Index pages only. That will be another ticket. :)

On Tue, Nov 3, 2015 at 12:24 AM, jdegraft <notifications@github.com> wrote:

> In omod/src/main/resources/webModuleApplicationContext.xml
> <https://github.com/openmrs/openmrs-module-legacyui/pull/14#discussion_r43684566>
> :
>
> >   			</props>
> >  		</property>
> >  	</bean>	
> >
> > + 	
> > +	
> > +	
> > +	<mvc:view-controller view-name="module/legacyui/admin/index" path="admin/index.htm"/>				
>
> One final question.
> Do I restrict these changes to index pages or in general?
> For instance if there is a reference to /admin/patientForm.jsp do I change
> it /module/legacyui/patientForm.jsp or leave that to be done in another
> ticket?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/openmrs/openmrs-module-legacyui/pull/14/files#r43684566>
> .
>



-- 
If we keep uppermost in our minds the unkind and unjust acts of others, we
shall find it impossible to love them as Christ has loved us; but if our
thoughts dwell upon the wondrous love and pity of Christ for us, the same
spirit will flow out to others.

kristopherschmidt: @jdegraft There are probably a few ways you can map URLs like /admin without making changes to the JSPs (btw if you do change links in the JSP they should reference /admin/index.htm). 
If you were to go back to the annotation method and re-introduce AdminPageController, you could write another method with @RequestMapping(value = "/admin") that returns "module/legacyui/admin/index".

Another thing you could do would be to put a mapping for /admin into the legacyUiUrlMapping bean and map it to an appropriate controller. The UrlFilenameViewController has a handy prefix and a suffix property, so if you were to set the prefix to "/module/legacyui" and the suffix to "/index", it looks like it would map a URL of /admin to /module/legacyui/admin/index. You could reuse that same controller instance in the legacyUiUrlMapping bean from /admin, /patients, etc, anything that has an 'index.jsp' underneath.

BTW r.e. view resolution -- the only reason /admin maps to /WEB-INF/view/admin.jsp right now is because there is no other explicit controller mapping in Spring, so Spring falls back to the 'springController' bean which is a UrlFilenameViewController (without a prefix or suffix) which interprets paths like /admin as a view named "admin", which then gets looked up by the internal view resolver as /WEB-INF/view/<viewname>.jsp. As soon as you add a mapping like /admin to another url handler mapping, it should work...

kristopherschmidt: @jdegraft maybe there is something that you are aware of that is special about /admin. I'm not quite sure why your initial suggestion of just mapping /admin wouldn't work, as in
<mvc:view-controller view-name="module/legacyui/admin/index" path="/admin"/>. If that doesn't, 
I think it's at least worth trying the controller with @RequestMapping(value="/admin")

jdegraft: @kristopherschmidt My understanding is mvc:controller generates an annotated controller so it is equivalent to the other controller approaches you have indicated. My comments were based on basically trying every approach and running it to ensure that it would not work.

jdegraft: @kristopherschmidt can you refer to my earlier comment as to what I thought was different about /admin and @wluyima 's response indicating that the problem couldn't be avoided by even using the servet api. (That was my understanding of his response anyway.)
Btw I haven't done any java in over 10 years, I have been mainly a .NET developer so this stuff is fairly new to me.

kristopherschmidt: @jdegraft I re-read both comments and I see what the problem is now. As @wluyima suggested, /admin is not even mapped to the Spring 'openmrs' DispatcherServlet in web.xml. 

However it is possible to map /admin to the DispatcherServlet (I did a quick test and it works). @wluyima why do you think the mapping would need to be removed if the legacyui module is stopped, any more than any other mapping like **/*.htm or **/*.form would need to be removed? If you stop the legacyui module those URLs will all just go back to looking for JSPs that don't exist, but that's ok because those URLs aren't supposed to work without legacyui?

wluyima: If you read the ticket description, what needs to be done is very clear, annotating a controller mapped to /admin doesn't work because the URL patterns mapped to the spring dispatcher servlet don't match URLs like /admin. a URL needs to end with in a certain form like <b>.list</b>, <b>.htm</b>, <b>.form</b> etc to be processed by the dispatcher servlett. The solution is to just change all these existing URLs to end with <b>index.htm</b>

jdegraft: @wluyima it is tested and working.
The approach did not allow mapping of "/admin" so my understanding was to change references in the UI from "/admin" to "/admin/index.htm".
Those changes have been tested and working.
Am I missing something?

kristopherschmidt: Your changes are fine, I'm pretty sure he was responding to me.

PR #17, opened 2015-11-11
LUI-17 Migrated PatientDashboard to LegacyUI Module

jdegraft:


omod/src/main/webapp/portlets/patientHeader.jsp
```diff
@@ -182,7 +182,7 @@
         <c:forEach items="${conceptIds}" var="conceptId">
             <td class="patientRecentObsConfigured">
                 <openmrs:concept conceptId="${conceptId}" var="c" nameVar="n" numericVar="num" shortestNameVar="sn">
-                    <span title="${n.description}">${sn}:</span>
+                    <span title="${n.name}">${sn}:</span>
```
wluyima: From the previous logic this was the description of the concept, so you need to replace this with the concept description instead of the name

PR #18, opened 2016-01-03
Fix for: Forgotten Password Form Leaks Valid Usernames [TRUNK-3933]

bholagabbar:
Prevented leakage of valid usernames. Incase of invalid username, a set of fake questions is asked whose answer is always false. Ticket link: https://issues.openmrs.org/browse/TRUNK-3933


bholagabbar:
Should I edit the same file or create a new request. The former would be much easier :)

dkayiwa:
@bholagabbar as long as you squash the commits into one. :)

PR #22, opened 2016-01-04
Fix for: Forgotten Password Form Leaks Valid Usernames [LUI-45]

bholagabbar:
Fixed the leakage of usernames in the legacy UI. If the user enters an invalid username, he will be assigned a random fake question on the basis of the hash-value of his username whose answer will always be false. The issue of no secret question/answer prompts an error. 
Added a few tests

### DIFF
--- a/omod/src/main/java/org/openmrs/web/WebComponentRegistrar.java
+++ b/omod/src/main/java/org/openmrs/web/WebComponentRegistrar.java
@@ -29,6 +29,9 @@ public class WebComponentRegistrar implements ServletContextAware {
 		
 		ServletRegistration openmrsServletReg = servletContext.getServletRegistration("openmrs");
 		addMappings(openmrsServletReg, "*.htm", "*.form", "*.list", "*.json", "*.field", "*.portlet");
+		
+		addMappings(servletContext.getServletRegistration("jsp"), "*.withjstl");
+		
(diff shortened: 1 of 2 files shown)


bholagabbar:
Sure. As the OpenMRS password goes, it should have atleast 1 uppercase and 1 lowercase letter. So I am trying to make sure that the condition is satisfied..

dkayiwa:
@bholagabbar the password validation rules should not be hard coded because they are determined by global properties like: "security.passwordRequiresUpperAndLowerCase", "security.passwordRequiresNonDigit", "security.passwordMinimumLength", and more. For instance the user can turn this off by setting the value of this global property to false.

bholagabbar:
What are the changes to make? Why does the test not work? I mean idk why it is passing on otherform.form....

bholagabbar:
I will work on them and initiate a new request

dkayiwa:
@bholagabbar can you step back a bit, do some googling and read about unit testing in general? :)

bholagabbar:
Ok will do that...

bholagabbar:
Ahhh I have got my mistake. I was trying and working with the test modules of the old open-mrs core. I have taken the idea from the changePasswordFormControllerTest.java in legacyui and will be basing my new tests upon it. Hope that's ok. And sorry for being a noob. I assure you I will learn everything about OpenMRS in a few weeks :)

PR #24, opened 2016-01-13
LUI-45. Fixed code. Unable to write tests

bholagabbar:
Pull request on the request of Wyclif Lumia.

Have a look at the ForgotPasswordFormControllerTest.java file where I try and execute the tests. The test fails and the error is:

<b>Failed tests: shouldNotNotFailOnNotFoundUsernameTEST(org.openmrs.web.controller.ForgotPasswordFormControllerTest): expected:<[check]> but was:<[What is your grandfather's home town?]</b>?

This means that the user is not being detected and a secret question is being randomly assigned. However, "bruno" is there in the test data. How do I overcome this?

### DIFF
--- a/omod/src/main/java/org/openmrs/web/controller/ForgotPasswordFormController.java
+++ b/omod/src/main/java/org/openmrs/web/controller/ForgotPasswordFormController.java
@@ -9,8 +9,11 @@
  */
 package org.openmrs.web.controller;
 
+import java.util.ArrayList;
+import java.util.Collections;
 import java.util.Date;
 import java.util.HashMap;
+import java.util.List;
 import java.util.Map;
 
 import javax.servlet.ServletException;
@@ -20,7 +23,6 @@ import javax.servlet.http.HttpSession;
 
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
-import org.apache.commons.lang.StringUtils;
 import org.openmrs.User;
 import org.openmrs.api.context.Context;
 import org.openmrs.util.PrivilegeConstants;
@@ -68,7 +70,7 @@ public class ForgotPasswordFormController extends SimpleFormController {
 	 *      org.springframework.validation.BindException)
 	 */
 	protected ModelAndView onSubmit(HttpServletRequest request, HttpServletResponse response, Object obj,
-	        BindException errors) throws Exception {
+	                                BindException errors) throws Exception {
 		
 		HttpSession httpSession = request.getSession();
 		
@@ -123,12 +125,20 @@ public class ForgotPasswordFormController extends SimpleFormController {
 				finally {
 					Context.removeProxyPrivilege(PrivilegeConstants.GET_USERS);
 				}
-				
-				if (user == null /*|| StringUtils.isEmpty(user.getSecretQuestion())*/) {
+
+				String secretQuestion = null;
+				if(user != null) {
+					secretQuestion = Context.getUserService().getSecretQuestion(user);
+				}
+
+				if (user == null) {
+					httpSession.setAttribute(WebConstants.OPENMRS_MSG_ATTR, "auth.question.fill");
+					request.setAttribute("secretQuestion", getRandomFakeSecretQuestion(username));
+				} else if (secretQuestion == null || secretQuestion.equals("")) {
 					httpSession.setAttribute(WebConstants.OPENMRS_ERROR_ATTR, "auth.question.empty");
 				} else {
 					httpSession.setAttribute(WebConstants.OPENMRS_MSG_ATTR, "auth.question.fill");
-					//request.setAttribute("secretQuestion", user.getSecretQuestion());
+					request.setAttribute("secretQuestion", secretQuestion);
 					
 					// reset the forgotPasswordAttempts because they have a right user.
 					// they will now have 5 more chances to get the question right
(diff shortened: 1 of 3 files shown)


wluyima:
Good practice is not to code off the master branch in your fork but to create a separate branch for each ticket your work on, otherwise you can end with commits for other tickets in the one pull request because github adds all commits made to a branch to the pull request linked to it

bholagabbar:
Ok..but can you please tell me what is wrong is the tests..I will take care of the rest in the future

omod/src/main/java/org/openmrs/web/controller/ForgotPasswordFormController.java
```diff
@@ -123,12 +121,20 @@ protected ModelAndView onSubmit(HttpServletRequest request, HttpServletResponse
...
 				}
-				
-				if (user == null /*|| StringUtils.isEmpty(user.getSecretQuestion())*/) {
+
+				String secretQuestion = null;
+				if(user != null) {
+					secretQuestion = Context.getUserService().getSecretQuestion(user);
+				}
+
+				if (user == null) {
```
wluyima: I think you merge these 2 if clauses because you're checking if user is not null above and here you're checking if they are null

omod/src/main/java/org/openmrs/web/controller/ForgotPasswordFormController.java
```diff
@@ -147,41 +153,98 @@ protected ModelAndView onSubmit(HttpServletRequest request, HttpServletResponse
 				finally {
 					Context.removeProxyPrivilege(PrivilegeConstants.GET_USERS);
 				}
+
+				String secretQuestion= null;
```
wluyima: Isn't this duplicate code to the if clause on line 125 above?

wluyima: Ooops! This is a different method, ignore my comment above

omod/src/main/java/org/openmrs/web/controller/ForgotPasswordFormController.java
```diff
@@ -147,41 +153,98 @@ protected ModelAndView onSubmit(HttpServletRequest request, HttpServletResponse
...
 					
 					StringBuilder randomPassword = new StringBuilder();
-					for (int i = 0; i < 8; i++) {
-						randomPassword.append(String.valueOf((Math.random() * (127 - 48) + 48)));
+					ArrayList<Character> buildRandomPassword = new ArrayList<Character>();
+
+					//Ensure atleast 1 uppercase, 1 lowercase as well as 1 digit in password (Otherwise, exception is thrown)
+					buildRandomPassword.add((char) (65 + (int) Math.abs(Math.random()) % 26));
+					buildRandomPassword.add((char) (97 + (int) Math.abs(Math.random()) % 26));
+					buildRandomPassword.add((char) (48 + (int) Math.abs(Math.random()) % 10));
```
wluyima: Why are you generating a random password?

omod/src/main/java/org/openmrs/web/controller/ForgotPasswordFormController.java
```diff
@@ -147,41 +153,98 @@ protected ModelAndView onSubmit(HttpServletRequest request, HttpServletResponse
...
+		User testUser = Context.getUserService().getUserByUsername(username);
+		String testSecretQuestion;
+		if (testUser != null) {
+			testSecretQuestion = Context.getUserService().getSecretQuestion(testUser);
+		} else {
+			testSecretQuestion = getRandomFakeSecretQuestion(username);
+		}
 		request.setAttribute("uname", username);
+		request.setAttribute("secretQuestion", testSecretQuestion);
+
```
bholagabbar: I have purposely rechecked for the getSecretQuestion. This was to reverify I was fetching it in the tests. Will remove it after tests work..

omod/src/main/java/org/openmrs/web/controller/ForgotPasswordFormController.java
wluyima: As in, is it part of this ticket to change this logic too?

omod/src/test/java/org/openmrs/web/controller/ForgotPasswordFormControllerTest.java
```diff
@@ -0,0 +1,68 @@
+/**
+ * This Source Code Form is subject to the terms of the Mozilla Public License,
```
bholagabbar: This file is where the problem is. If this starts detecting my data, this issue is solved as I will generate relevant tests and cleanup code ASAP.

omod/src/main/java/org/openmrs/web/controller/ForgotPasswordFormController.java
```diff
@@ -147,41 +153,98 @@ protected ModelAndView onSubmit(HttpServletRequest request, HttpServletResponse
...
 		return showForm(request, response, errors);
 	}
 	
+	public String getRandomFakeSecretQuestion(String username) {
+
+		List<String> questions = new ArrayList<String>();
+		
+		questions.add(Context.getMessageSourceService().getMessage("What is your best friend's name?"));
```
wluyima: These need to be message codes that are defined in the messages.properties file with these as the english translations

omod/src/test/java/org/openmrs/web/controller/ForgotPasswordFormControllerTest.java
```diff
@@ -0,0 +1,68 @@
...
+	protected static final String TEST_DATA= "org/openmrs/web/controller/include/ForgotPasswordFormControllerTest.xml";
+
+	/**
+	 * Check to see if the admin's secret question comes back
+	 *
+	 * @throws Exception
+	 */
+	@Test
+	public void shouldNotNotFailOnNotFoundUsernameTEST() throws Exception {
+		initializeInMemoryDatabase();
```
wluyima: This is already done for you by the super class you extended

omod/src/test/java/org/openmrs/web/controller/ForgotPasswordFormControllerTest.java
```diff
@@ -0,0 +1,68 @@
...
+		executeDataSet(TEST_DATA);
+		authenticate();
```
wluyima: This too is already done by the super class

omod/src/test/java/org/openmrs/web/controller/ForgotPasswordFormControllerTest.java
```diff
@@ -0,0 +1,68 @@
...
+
+		ForgotPasswordFormController controller = new ForgotPasswordFormController();
```
wluyima: ForgotPasswordFormController is not actually an annotated controller, you need to fetch an instance of it from the application context, see [this](https://github.com/openmrs/openmrs-module-legacyui/blob/master/omod/src/test/java/org/openmrs/web/controller/ConceptFormControllerTest.java#L87) for an example.

omod/src/test/java/org/openmrs/web/controller/ForgotPasswordFormControllerTest.java
wluyima: If you wish to use your own dataset, you would have to override useInMemoryDatabase and return false then you can keep lines 52 and 54

omod/src/test/resources/org/openmrs/web/controller/include/ForgotPasswordFormControllerTest
```diff
@@ -0,0 +1,16 @@
...
+    v. 2.0. If a copy of the MPL was not distributed with this file, You can
+    obtain one at http://mozilla.org/MPL/2.0/. OpenMRS is also distributed under
+    the terms of the Healthcare Disclaimer located at http://openmrs.org/license.
+
+    Copyright (C) OpenMRS Inc. OpenMRS is a registered trademark and the OpenMRS
+    graphic logo is a trademark of OpenMRS Inc.
+
+-->
+<dataset>
+    <users user_id="501" person_id="501" system_id="2-6" username="bruno" password="ff6d1655327d385e11a04e1632b4f33ceb8fafd2" salt="5d32c5fd9fde391c755f1f4dfd5e1d6e3debe6" secret_question="what is your name bruno?" creator="1" date_created="2008-08-15 15:46:47.0" changed_by="1" date_changed="2008-08-15 15:47:07.0" retired="true" retire_reason="Test purposes" uuid="c1d8f5c2-e131-11de-babe-001e378eb67e"/>
```
wluyima: It could be that these datasets are overriding each other, see the comments above.

omod/src/test/java/org/openmrs/web/controller/ForgotPasswordFormControllerTest.java
```diff
@@ -0,0 +1,68 @@
...
+import java.net.BindException;
+/**
+ * Test the different aspects of
+ * {@link org.openmrs.web.controller.ForgotPasswordFormController}
+ */
+public class ForgotPasswordFormControllerTest extends BaseModuleWebContextSensitiveTest {
+
+	private static Level initialLogLevel;
+
+	private static final String CLASS_TO_LOG = "org.openmrs.api.db.hibernate.HibernateUserDAO";
```
wluyima: This should be:
 private static Log log = LogFactory.getLog(ForgotPasswordFormControllerTest.class);

omod/src/main/java/org/openmrs/web/controller/ForgotPasswordFormController.java
bholagabbar: Yes It is.

omod/src/test/java/org/openmrs/web/controller/ForgotPasswordFormControllerTest.java
bholagabbar: What dataset should I be using? Because I have tried everything and still not detecting..

omod/src/test/java/org/openmrs/web/controller/ForgotPasswordFormControllerTest.java
bholagabbar: If I use `ForgotPasswordFormController controller = (ForgotPasswordFormController) applicationContext.getBean("conceptForm");`
<br>
Then the error is:<br>
`java.lang.ClassCastException: org.openmrs.web.controller.ConceptFormController cannot be cast to org.openmrs.web.controller.ForgotPasswordFormController
	at org.openmrs.web.controller.ForgotPasswordFormControllerTest.shouldNotNotFailOnNotFoundUsernameTEST(ForgotPasswordFormControllerTest.java:57)`

omod/src/test/resources/org/openmrs/web/controller/include/ForgotPasswordFormControllerTest
bholagabbar: As I asked previously, what dataset should I be using? Would it be possible for you to give me a template of the correct test methods that would work? That would be really helpful.

omod/src/test/java/org/openmrs/web/controller/ForgotPasswordFormControllerTest.java
bholagabbar: What exactly is this for? To log the tests? Or the dataset? I'm sorry I don't understand what this does

omod/src/main/java/org/openmrs/web/controller/ForgotPasswordFormController.java
bholagabbar: The previous code that generated passwords was flawed and it threw me and exception as it did not take care of caps and numbers. I have refined that part of the code. You can check using diff

omod/src/main/java/org/openmrs/web/controller/ForgotPasswordFormController.java
bholagabbar: Ok will do

omod/src/main/java/org/openmrs/web/controller/ForgotPasswordFormController.java
wluyima: I think you could use apache common's RandomStringUtils to generate a random alphanumeric string and then just add a random upper case letter to it, RandomStringUtils still can get you this letter

omod/src/test/java/org/openmrs/web/controller/ForgotPasswordFormControllerTest.java
wluyima: I also don't know why you added it, I guess If you're not using anywhere in your code of course you can omit it

omod/src/test/java/org/openmrs/web/controller/ForgotPasswordFormControllerTest.java
wluyima: Either works, just pick one. I would just get rid of the rest dataset and lines 52, 53 and 54. Then set the secret question for the test user within the test as the first thing

omod/src/test/java/org/openmrs/web/controller/ForgotPasswordFormControllerTest.java
wluyima: Because you forgot to change the bean name to fetch from the application context, in fact the error message kind of gives a clue because you are fetching a concept form but expecting to get back a forgot password form

omod/src/test/resources/org/openmrs/web/controller/include/ForgotPasswordFormControllerTest
wluyima: See my comment above, i wouldn't add a new dataset personally.

PR #25, opened 2016-01-14
Pull request to correct dysfunctional tests

bholagabbar:
Dkayiwa: Hope you can figure out the problems locally!

### DIFF
--- a/omod/src/main/java/org/openmrs/web/attribute/handler/ConceptFieldGenDatatypeHandler.java
+++ b/omod/src/main/java/org/openmrs/web/attribute/handler/ConceptFieldGenDatatypeHandler.java
@@ -1,0 +1,64 @@
+/**
+ * This Source Code Form is subject to the terms of the Mozilla Public License,
+ * v. 2.0. If a copy of the MPL was not distributed with this file, You can
+ * obtain one at http://mozilla.org/MPL/2.0/. OpenMRS is also distributed under
+ * the terms of the Healthcare Disclaimer located at http://openmrs.org/license.
+ *
+ * Copyright (C) OpenMRS Inc. OpenMRS is a registered trademark and the OpenMRS
+ * graphic logo is a trademark of OpenMRS Inc.
+ */
+package org.openmrs.web.attribute.handler;
+
+import java.util.Map;
+
+import org.openmrs.Concept;
+import org.openmrs.customdatatype.CustomDatatype;
+import org.openmrs.customdatatype.datatype.ConceptDatatype;
+import org.springframework.stereotype.Component;
+
+/**
+ * Handler for the Concept custom datatype
+ */
+@Component
+public class ConceptFieldGenDatatypeHandler extends SerializingFieldGenDatatypeHandler<ConceptDatatype, Concept> {
+	
+	/**
+	 * @see SerializingFieldGenDatatypeHandler#getWidgetName()
+	 */
+	@Override
+	public String getWidgetName() {
+		return "org.openmrs.Concept";
+	}
+	
+	/**
+	 * @see SerializingFieldGenDatatypeHandler#toHtml(CustomDatatype, String)
+	 */
+	@Override
+	public String toHtml(CustomDatatype datatype, String valueReference) {
+		return toHtmlSummary(datatype, valueReference).getSummary();
+	}
+	
+	/**
+	 * @see SerializingFieldGenDatatypeHandler#toHtmlSummary(CustomDatatype, String)
+	 */
+	@Override
+	public CustomDatatype.Summary toHtmlSummary(CustomDatatype datatype, String valueReference) {
+		return datatype.getTextSummary(valueReference);
+	}
+	
+	/**
+	 * @see SerializingFieldGenDatatypeHandler#getWidgetConfiguration()
+	 */
+	@Override
+	public Map<String, Object> getWidgetConfiguration() {
+		return null;
+	}
+	
+	/**
+	 * @see SerializingFieldGenDatatypeHandler#setHandlerConfiguration(String)
+	 */
+	@Override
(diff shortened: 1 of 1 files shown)


omod/src/main/java/org/openmrs/web/controller/ForgotPasswordFormController.java
```diff
@@ -123,12 +125,20 @@ protected ModelAndView onSubmit(HttpServletRequest request, HttpServletResponse
...
 				}
-				
-				if (user == null /*|| StringUtils.isEmpty(user.getSecretQuestion())*/) {
+
+				String secretQuestion = null;
+				if(user != null) {
+					secretQuestion = Context.getUserService().getSecretQuestion(user);
+				}
+
+				if (user == null) {
```
wluyima: You should probably merge these if clauses

omod/src/main/java/org/openmrs/web/controller/ForgotPasswordFormController.java
```diff
@@ -147,41 +157,103 @@ protected ModelAndView onSubmit(HttpServletRequest request, HttpServletResponse
...
-				} else if (/*user.getSecretQuestion() != null &&*/ Context.getUserService().isSecretAnswer(user, secretAnswer)) {
+				} else if (secretQuestion != null && Context.getUserService().isSecretAnswer(user, secretAnswer)) {
 					
 					StringBuilder randomPassword = new StringBuilder();
-					for (int i = 0; i < 8; i++) {
-						randomPassword.append(String.valueOf((Math.random() * (127 - 48) + 48)));
+					ArrayList<Character> buildRandomPassword = new ArrayList<Character>();
+
+					//Ensure atleast 1 uppercase, 1 lowercase as well as 1 digit in password (Otherwise, exception is thrown)
+					buildRandomPassword.add((char) (65 + (int) Math.abs(Math.random()) % 26));
```
wluyima: You can use RandomStringUtils to generate an alphanumeric string and an upper case letter and concatenate them instead of doing this yourself

omod/src/test/java/org/openmrs/web/controller/ForgotPasswordFormControllerTest.java
```diff
@@ -0,0 +1,56 @@
...
+ * Test the different aspects of
+ * {@link org.openmrs.web.controller.ForgotPasswordFormController}
+ */
+public class ForgotPasswordFormControllerTest extends BaseModuleWebContextSensitiveTest {
+
+	protected static final String TEST_DATA= "org/openmrs/web/controller/include/ForgotPasswordFormControllerTest.xml";
+
+	@Before
+	public void updateSearchIndex() {
+		super.updateSearchIndex();
```
wluyima: Why do you need this? This is only required for concepts

omod/src/test/java/org/openmrs/web/controller/ForgotPasswordFormControllerTest.java
```diff
@@ -0,0 +1,56 @@
...
+	}
+
+	/**
+	 * Check to see if the admin's secret question comes back
+	 *
+	 * @throws Exception
+	 */
+	@Test
+	public void shouldNotNotFailOnNotFoundUsernameTEST() throws Exception {
+		executeDataSet(TEST_DATA);
```
wluyima: If your test still fails, it could be because you are adding this test data which i think duplicates the standard dataset

wluyima: I suggest you get rid of it and set the secret question for one of the users in the standard test dataset and use that before the test

bholagabbar: Ok. How do I include the standard dataset? Is it already inherited?

omod/src/test/java/org/openmrs/web/controller/ForgotPasswordFormControllerTest.java
bholagabbar: Ok I was just trying out new things..seeing if it works

omod/src/main/java/org/openmrs/web/controller/ForgotPasswordFormController.java
bholagabbar: Yes, but again, that would involve some conversion etc...I will do it if you say so

omod/src/main/java/org/openmrs/web/controller/ForgotPasswordFormController.java
bholagabbar: yes I will do that. Sorry you have had to repeat yourself. I was just focusing on tests

omod/src/test/java/org/openmrs/web/controller/ForgotPasswordFormControllerTest.java
wluyima: It's include by default, so you don't have to do anything

wluyima: As long as your test is a subclass of BaseContextSensitiveTest which BaseModuleWebContextSensitiveTest is and in turn is your test class is already

omod/src/main/java/org/openmrs/web/controller/ForgotPasswordFormController.java
wluyima: What conversion? All you have to is:

`RandomStringUtils.randomAlphabetic(1).toUpperCase() + RandomStringUtils.randomAlphanumeric(minChar)`

minChar should be the value of the GP_PASSWORD_MINIMUM_LENGTH global property value, default value should be 7(technically 8) but the eighth letter is the random uppercase letter we are prepending to it

PR #28, opened 2016-01-19
Final fix with tests for LUI-45 [Forgot Password Form Leaks Valid Usernames]

bholagabbar:

### DIFF
--- a/omod/pom.xml
+++ b/omod/pom.xml
@@ -96,6 +96,43 @@
             <version>3.0.1</version>
             <scope>provided</scope>
         </dependency>
+        
+        <dependency>
+			<groupId>javax.servlet</groupId>
+			<artifactId>jsp-api</artifactId>
+			<version>2.0</version>
+		</dependency>
+		<dependency>
+			<groupId>javax.servlet</groupId>
+			<artifactId>jstl</artifactId>
+			<version>1.1.2</version>
+		</dependency>
+		<dependency>
+			<groupId>taglibs</groupId>
+			<artifactId>request</artifactId>
+			<version>1.0.1</version>
+		</dependency>
+		<dependency>
+			<groupId>taglibs</groupId>
+			<artifactId>response</artifactId>
+			<version>1.0.1</version>
+		</dependency>
+		<dependency>
+			<groupId>taglibs</groupId>
+			<artifactId>standard</artifactId>
+			<version>1.1.2</version>
+		</dependency>
+		<dependency>
+			<groupId>taglibs</groupId>
+			<artifactId>page</artifactId>
+		<version>1.0.1</version>
+			</dependency>
+		
+		<dependency>
+			<groupId>jfree</groupId>
+			<artifactId>jfreechart</artifactId>
+				<version>1.0.12</version>
+		</dependency>
 
 	</dependencies>
 


omod/src/main/java/org/openmrs/web/controller/ForgotPasswordFormController.java
```diff
@@ -119,16 +121,23 @@ protected ModelAndView onSubmit(HttpServletRequest request, HttpServletResponse
...
 					Context.removeProxyPrivilege(PrivilegeConstants.GET_USERS);
 				}
 				
-				if (user == null /*|| StringUtils.isEmpty(user.getSecretQuestion())*/) {
+				String secretQuestion = null;
+				if (user != null) {
+					secretQuestion = Context.getUserService().getSecretQuestion(user);
+				}
+				
+				if (user == null) {
```
wluyima: Can you please merge these 2 if clauses?

omod/src/main/java/org/openmrs/web/controller/ForgotPasswordFormController.java
```diff
@@ -143,26 +152,46 @@ protected ModelAndView onSubmit(HttpServletRequest request, HttpServletResponse
...
 				}
 				
+				String secretQuestion = null;
+				if (user != null) {
+					secretQuestion = Context.getUserService().getSecretQuestion(user);
+				}
+				
 				// check the secret question again in case the user got here "illegally"
-				if (user == null /*|| StringUtils.isEmpty(user.getSecretQuestion())*/) {
+				if (user == null) {
```
wluyima: Same with this, merge the if clauses to make this more readable

omod/src/main/java/org/openmrs/web/controller/ForgotPasswordFormController.java
```diff
@@ -143,26 +152,46 @@ protected ModelAndView onSubmit(HttpServletRequest request, HttpServletResponse
...
+					buildRandomPassword.add((char) (48 + (int) Math.abs(Math.random()) % 10));
+					//Build rest of the password
+					for (int i = 3; i < 8; i++) {
+						buildRandomPassword.add((char) (Math.random() * (127 - 48) + 48));
+					}
+					//Shuffle it
+					Collections.shuffle(buildRandomPassword);
+					//Finally, append it to the string
+					for (char characterInPassword : buildRandomPassword) {
+						randomPassword.append(characterInPassword);
```
bholagabbar: Before you ask me to change the password code, I tried doing it with RandomStringUtils but I kept getting an error. I think I'll stick to this. Seems to work out fine. I have tested it

omod/src/main/java/org/openmrs/web/controller/ForgotPasswordFormController.java
```diff
@@ -143,26 +152,46 @@ protected ModelAndView onSubmit(HttpServletRequest request, HttpServletResponse
...
-				} else if (/*user.getSecretQuestion() != null &&*/ Context.getUserService().isSecretAnswer(user, secretAnswer)) {
+				} else if (secretQuestion != null && Context.getUserService().isSecretAnswer(user, secretAnswer)) {
 					
 					StringBuilder randomPassword = new StringBuilder();
-					for (int i = 0; i < 8; i++) {
-						randomPassword.append(String.valueOf((Math.random() * (127 - 48) + 48)));
+					ArrayList<Character> buildRandomPassword = new ArrayList<Character>();
+					
+					//Ensure atleast 1 uppercase, 1 lowercase as well as 1 digit in password (Otherwise, exception is thrown)
+					buildRandomPassword.add((char) (65 + (int) Math.abs(Math.random()) % 26));
```
wluyima: Can you use RandomStringUtils for these?

omod/src/main/java/org/openmrs/web/controller/ForgotPasswordFormController.java
bholagabbar: Yup I have done it. I'll incorporate all the changes you ask me to and create a new pull request asap

omod/src/test/java/org/openmrs/web/controller/ForgotPasswordFormControllerTest.java
```diff
@@ -0,0 +1,281 @@
...
+/**
+ * Test the different aspects of
+ * {@link org.openmrs.web.controller.ForgotPasswordFormController}
+ */
+
+public class ForgotPasswordFormControllerTest extends BaseModuleWebContextSensitiveTest {
+	
+	protected static final String TEST_DATA = "org/openmrs/web/controller/include/ForgotPasswordFormControllerTest.xml";
+	
+	@Before
```
wluyima: We don't recommend this, if you really have to then make a call to the superclass' method first

omod/src/test/java/org/openmrs/web/controller/ForgotPasswordFormControllerTest.java
```diff
@@ -0,0 +1,281 @@
...
+	public void runBeforeEachTest() throws Exception {
+		executeDataSet(TEST_DATA);
+		Context.logout();
+	}
+	
+	/**
+	 * Log out the current user after all unit test cases just in case the password was reset
+	 */
+	@After
```
wluyima: I don't think this is necessary, you can get rod of it

omod/src/test/java/org/openmrs/web/controller/ForgotPasswordFormControllerTest.java
wluyima: Otherwise you don't need this

omod/src/main/java/org/openmrs/web/controller/ForgotPasswordFormController.java
bholagabbar: Yeah as I mentioned, I kept getting an error. Idk why. Somehow, RandomStringUtil.AlphaNumeric did not generate a vaild password. If you wish, I can create a new method called 'createRandomPassword' which will contain this code. This code is fool proof generation of password :)

omod/src/test/java/org/openmrs/web/controller/ForgotPasswordFormControllerTest.java
```diff
@@ -0,0 +1,281 @@
...
+	}
+	
+	/**
+	 * Check to see if the admin's secret question comes back
+	 *
+	 * @throws Exception
+	 */
+	
+	public void shouldNotNotFailOnNotFoundUsername() throws Exception {
+
```
wluyima: It would be nice to add an assertion here at the start that there is no user actually with the username

omod/src/test/java/org/openmrs/web/controller/ForgotPasswordFormControllerTest.java
```diff
@@ -0,0 +1,281 @@
...
+		
+		ForgotPasswordFormController controller = new ForgotPasswordFormController();
+		
+		
+		MockHttpServletRequest request = new MockHttpServletRequest();
+		MockHttpServletResponse response = new MockHttpServletResponse();
+		
+		
+		request.setParameter("uname", "bruno");
+		request.setParameter("secretAnswer", "valid secret answer");
```
wluyima: I thought you only provide the user name, if it is not valid it should return the form with a random secret question

omod/src/test/java/org/openmrs/web/controller/ForgotPasswordFormControllerTest.java
bholagabbar: I'm not sure what you're saying. I have asserted the user here...see the first assert

omod/src/test/java/org/openmrs/web/controller/ForgotPasswordFormControllerTest.java
```diff
@@ -0,0 +1,281 @@
...
+	}
+	
+	/**
+	 * If a user enters the wrong secret answer, they should be kicked back to the form and not be
+	 * authenticated
+	 *
+	 * @throws Exception
+	 */
+	@Test
+	public void shouldNotAuthenticateAsUserWithValidSecretQuestion() throws Exception {
```
wluyima: With a valid or invalid secret question?

omod/src/test/java/org/openmrs/web/controller/ForgotPasswordFormControllerTest.java
```diff
@@ -0,0 +1,281 @@
...
+		}
+		
+		// those were the first five, now the sixth request (with a valid user) should fail
+		MockHttpServletRequest request = new MockHttpServletRequest();
+		request.setMethod("POST");
+		
+		request.addParameter("uname", "validuser");
+		
+		controller.handleRequest(request, new MockHttpServletResponse());
+		Assert.assertNull(request.getAttribute("secretQuestion"));
```
wluyima: The proper way to test that a user is locked out is to check that the user property that indicates a user is locked out is set, i think a value is supposed to be the timestamp of when the lockout happened

omod/src/test/java/org/openmrs/web/controller/ForgotPasswordFormControllerTest.java
```diff
@@ -0,0 +1,281 @@
...
+	}
+	
+	/**
+	 * If a user enters 4 username requests, the 5th one should reset the lockout and they should be
+	 * allowed 5 attempts at the secret answer
+	 *
+	 * @throws Exception
+	 */
+	@Test
+	public void shouldGiveUserFiveSecretAnswerAttemptsAfterLessThanFiveFailedUsernameAttempts() throws Exception {
```
wluyima: I think you can phrase this better, i found it hard to understand until i read the javadocs above

omod/src/test/java/org/openmrs/web/controller/ForgotPasswordFormControllerTest.java
```diff
@@ -0,0 +1,281 @@
...
+		
+		request.addParameter("uname", "validuser");
+		
+		controller.handleRequest(request, new MockHttpServletResponse());
+		
+		Assert.assertNotNull(request.getAttribute("secretQuestion"));
+		
+		// now the user has 5 chances at the secret answer
+		
+		// fifth request
```
wluyima: Couldn't you put these too in a for loop?

omod/src/test/java/org/openmrs/web/controller/ForgotPasswordFormControllerTest.java
```diff
@@ -0,0 +1,281 @@
...
+		
+		request7.addParameter("uname", "validuser");
+		request7.addParameter("secretAnswer", "valid secret answer");
+		controller.handleRequest(request7, new MockHttpServletResponse());
+		
+		Assert.assertTrue(Context.isAuthenticated());
+	}
+	
+	@Test
+	public void shouldAuthenticateWithInValidSecretQuestionIfUserIsNull() throws Exception {
```
wluyima: This seems wrong to me, why would you authenticate a request with an invalod secret question and null user? This should fail

omod/src/test/java/org/openmrs/web/controller/ForgotPasswordFormControllerTest.java
bholagabbar: Yeah It should be Invalid. I'll change the user to 'invaliduser' and hence the invalid question will be assigned and that will not authenticate him. Is that alright?

omod/src/test/java/org/openmrs/web/controller/ForgotPasswordFormControllerTest.java
bholagabbar: Yeah my bad...I'll make the change. Was working with 1-2 old codes of mine that's why

omod/src/test/java/org/openmrs/web/controller/ForgotPasswordFormControllerTest.java
bholagabbar: Oh sorry. It should have been 'shouldNotAuthenticate' I'll change it

omod/src/test/java/org/openmrs/web/controller/ForgotPasswordFormControllerTest.java
bholagabbar: This seems much more lucid. The 5th request is the final chance. I think I would let this be this way

omod/src/test/java/org/openmrs/web/controller/ForgotPasswordFormControllerTest.java
bholagabbar: I think this also make sense...there will be no secret question assigned to him. This happens in no other case..

bholagabbar:
I've made most of the changes. Do you agree with my comments?

omod/src/main/java/org/openmrs/web/controller/ForgotPasswordFormController.java
wluyima: I added a code snippet of what you need to use to create a valid and random password on one of the pull requests you closed. You need to use a combination of RandomStringUtils.alphaNumeric and RandomStringUtils.random and convert the returned letter to upper case and pre-append it to the alpha numeric string , the length would need to depend on the value of the global property for password minimum length

omod/src/main/java/org/openmrs/web/controller/ForgotPasswordFormController.java
wluyima: What exception are you getting? I'm on IRC

omod/src/test/java/org/openmrs/web/controller/ForgotPasswordFormControllerTest.java
wluyima: At the start of the test method you check that UserService.getUserByUsername("bruno") returns null to gurantee that the user doesn't actually exist, otherwise if some one changed the test dataset and tampered with the user name it can mess up this test

omod/src/test/java/org/openmrs/web/controller/ForgotPasswordFormControllerTest.java
wluyima: I thought you don't authenticate with the secret question, providing an username results in a random secret question

omod/src/test/java/org/openmrs/web/controller/ForgotPasswordFormControllerTest.java
wluyima: You can have a secret question when you are not locked out, if you're saying the user is locked in the test method name out then you need to check that they are actually locked out in the database, it's importing for test method name to be reflective of the actual thing you are testing

omod/src/test/java/org/openmrs/web/controller/ForgotPasswordFormControllerTest.java
wluyima: I meant to put request 5 and 6 to avoid duplication of code but it's okay if you wish to leave it this way

wluyima:
You don't need to create a new pull request for every commit

wluyima:
as long as you’re working on the same ticket, just commit and push to the same branch, github will automatically update the pull request as long as it is not yet merged

PR #32, opened 2016-01-20
Final fix for LUI-45 [Forgot Password Form Leaks Valid Usernames]

bholagabbar:

### DIFF
--- a/omod/src/test/java/org/openmrs/web/controller/ConceptFormControllerTest.java
+++ b/omod/src/test/java/org/openmrs/web/controller/ConceptFormControllerTest.java
@@ -58,6 +58,7 @@ import org.springframework.web.servlet.ModelAndView;
 /**
  * Unit testing for the ConceptFormController.
  */
+@Ignore
 public class ConceptFormControllerTest extends BaseModuleWebContextSensitiveTest {
 	
 	@Autowired


omod/src/main/java/org/openmrs/web/controller/ForgotPasswordFormController.java
```diff
@@ -143,45 +150,79 @@ protected ModelAndView onSubmit(HttpServletRequest request, HttpServletResponse
...
 		loginAttemptsByIP.put(ipAddress, forgotPasswordAttempts);
+		
 		request.setAttribute("uname", username);
+		
 		return showForm(request, response, errors);
 	}
 	
+	public String getRandomPassword() {
+		
+		String randomString = RandomStringUtils.randomAlphabetic(1).toUpperCase() + RandomStringUtils.randomAlphanumeric(8)+ RandomStringUtils.randomNumeric(1);
```
wluyima: - This assignment is redundant since you never use the variable anywhere except to return it, just have a single return statement
- The count you pass to randomAlphanumeric() should be the value of the Global property OpenmrsConstants.GP_PASSWORD_MINIMUM_LENGTH, if it is blank, that's when you can default to 8
- Remove the randomNumeric at the end because the value returned by randomAlphanumeric already contains a number anyways

omod/src/test/java/org/openmrs/web/controller/ForgotPasswordFormControllerTest.java
```diff
@@ -0,0 +1,268 @@
...
+/**
+ * Test the different aspects of
+ * {@link org.openmrs.web.controller.ForgotPasswordFormController}
+ */
+
+public class ForgotPasswordFormControllerTest extends BaseModuleWebContextSensitiveTest {
+	
+	protected static final String TEST_DATA = "org/openmrs/web/controller/include/ForgotPasswordFormControllerTest.xml";
+	
+	@Before
```
wluyima: Why do you need to logout? I think it is redundant

bholagabbar:
Is that all?

omod/src/test/java/org/openmrs/web/controller/ForgotPasswordFormControllerTest.java
```diff
@@ -0,0 +1,268 @@
...
+	}
+	
+	/**
+	 * Check to see if the admin's secret question comes back
+	 *
+	 * @throws Exception
+	 */
+	
+	@Test
+	public void shouldNotNotFailOnNotFoundUsername() throws Exception {
```
wluyima: You have double 'Not' in this method name

omod/src/test/java/org/openmrs/web/controller/ForgotPasswordFormControllerTest.java
```diff
@@ -0,0 +1,268 @@
...
+		
+		ForgotPasswordFormController controller = new ForgotPasswordFormController();
+		
+		
+		MockHttpServletRequest request = new MockHttpServletRequest();
+		MockHttpServletResponse response = new MockHttpServletResponse();
+		
+		request.setParameter("uname", "validuser");
```
wluyima: - This test is a little confusing to me given the method name, why is this saying 'validuser'  when the method says shouldNotNotFailOn**NotFoundUsername**
- I would also assume if the username is not found, then the returned secret question should be a random fake one, i don't see where you're testing it within this test

omod/src/main/java/org/openmrs/web/controller/ForgotPasswordFormController.java
bholagabbar: In the OpenmrsConstants.class file, the value is GP_PASSWORD_MINIMUM_LENGTH = "security.passwordMinimumLength"; This returns a string. How do I use this? Should I add a comment and hardcode it as 8?

omod/src/test/java/org/openmrs/web/controller/ForgotPasswordFormControllerTest.java
```diff
@@ -0,0 +1,268 @@
...
+		
+		controller.handleRequest(request, response);
+		
+		Assert.assertEquals("validuser", request.getAttribute("uname"));
+		Assert.assertEquals("valid secret question", request.getAttribute("secretQuestion"));
+	}
+	
+	
+	@Test
+	public void shouldAuthenticateAsUserWithValidSecretQuestion() throws Exception {
```
wluyima: I'm also a little confused by this test, why does this test say 'Authenticate', I thought the forgot password is for recovering the password and not authenticating, i would assume when the user provides the correct username and secret question then you should test that they are shown the changePassword form, controller.handleRequest(request, response)  returns a ModelAndView object, you should be testing that the value returned by ModelAndView.getViewName() is the changePasswordForm, see [FormFormControllerTest](https://github.com/openmrs/openmrs-module-legacyui/blob/master/omod/src/test/java/org/openmrs/web/controller/form/FormFormControllerTest.java) for an example

omod/src/test/java/org/openmrs/web/controller/ForgotPasswordFormControllerTest.java
```diff
@@ -0,0 +1,268 @@
...
+	}
+	
+	/**
+	 * If a user enters the wrong secret answer, they should be kicked back to the form and not be
+	 * authenticated
+	 *
+	 * @throws Exception
+	 */
+	@Test
+	public void shouldNotAuthenticateAsUserWithInvalidSecretQuestion() throws Exception {
```
wluyima: I think you need to use some other term and not authenticate because they are not authenticating, how about 
**shouldRejectAUserWithInvalidSecretQuestion**

omod/src/test/java/org/openmrs/web/controller/ForgotPasswordFormControllerTest.java
bholagabbar: The name is a little confusing. What the method actually does that if the correct answer is applied, the user is logged in automatically and the admin's change password form in shown where he can change his password. That is what this test is checking. I can rename it if you wish

omod/src/test/java/org/openmrs/web/controller/ForgotPasswordFormControllerTest.java
```diff
@@ -0,0 +1,268 @@
...
+		ForgotPasswordFormController controller = (ForgotPasswordFormController) applicationContext.getBean("forgotPasswordForm");
+		
+		MockHttpServletRequest request = new MockHttpServletRequest();
+		request.setMethod("POST");
+		
+		request.addParameter("uname", "invaliduser");
+		request.addParameter("secretAnswer", "invalid secret answer");
+		
+		HttpServletResponse response = new MockHttpServletResponse();
+		controller.handleRequest(request, response);
```
wluyima: This returns a ModelAndView object, you need to test that it has the correct viewName

omod/src/main/java/org/openmrs/web/controller/ForgotPasswordFormController.java
bholagabbar: If I remove the number as you said, there is a good chance that the random.Alphanumeric will contain only aphabets. I got on of this errors on my tests 'org.openmrs.api.InvalidCharactersPasswordException: error.password.requireNumber'. Now you know why I had let that extra number be :)

bholagabbar: Here is the complete error log if you don't believe me :) http://pastebin.com/QiKNULCj

omod/src/test/java/org/openmrs/web/controller/ForgotPasswordFormControllerTest.java
bholagabbar: The tests fail if I do not logout :) I'm telling you, most of this code is tried and tested.

Here is the error log after removing Context.logout(): http://pastebin.com/bUaCpqiP

After adding is, all tests pass. You can check locally if you wish :)

omod/src/main/java/org/openmrs/web/controller/ForgotPasswordFormController.java
wluyima: - I have never run into that with alphaNumeric, alphaNumeric means number and letter, unless the documentation for that method is wrong which is misleading.
- OpenmrsConstants.GP_PASSWORD_MINIMUM_LENGTH is just the global property name, need to fetch the value of the global property using the AdministrationService

wluyima: You're right, RandomStringUtils.randomAlphanumeric can return just letter which is strange because the javadocs don't say so, you can leave the random number at the end

bholagabbar: I'm telling you the tests are failing! How else can I convince you? I have even pasted the error log of the tests!

omod/src/test/java/org/openmrs/web/controller/ForgotPasswordFormControllerTest.java
bholagabbar: OK I will check if the secret question returned is one of the fake secret question

omod/src/test/java/org/openmrs/web/controller/ForgotPasswordFormControllerTest.java
wluyima: I see what you're saying and you're correct based on the current code, but the thing is that we need to change that behvaior, the workflow should be such that when one provides the correct answer and username, they get taken to the changePassword form, the system should NEVER change the password for a user behind the scenes

omod/src/main/java/org/openmrs/web/controller/ForgotPasswordFormController.java
wluyima: See my comment above

omod/src/test/java/org/openmrs/web/controller/ForgotPasswordFormControllerTest.java
wluyima: I see why you're logging out, what am saying is that the reason why you need to logout is because the code processing the request in the controller is doing something wrong, it is authenticating the user whose username and answer you provide when you submit the form which is wrong, it should just take the user to the changePassword form. In theory, i think we should not be generating a random password to set as the user password

omod/src/test/java/org/openmrs/web/controller/ForgotPasswordFormControllerTest.java
bholagabbar: Nope the system in not changing it. This password is assigned to log him in. Basically the 'Old password' section is filled up. He has to enter the new password. That's it

omod/src/test/java/org/openmrs/web/controller/ForgotPasswordFormControllerTest.java
bholagabbar: It returns null..If I get the value of the ModelView.getView().toString. I'm not sure how to user the Assert statement here.

omod/src/main/java/org/openmrs/web/controller/ForgotPasswordFormController.java
bholagabbar: `Integer.parseInt(Context.getAdministrationService().getGlobalProperty(OpenmrsConstants.GP_PASSWORD_MINIMUM_LENGTH))` **returns null**<br>
I'll stick to 8 hardcoded with a comment?

omod/src/test/java/org/openmrs/web/controller/ForgotPasswordFormControllerTest.java
wluyima: If you look at the code closely there is commented out line that changes the password, we just commented it recently when we moved that controller from to the module from core, we need to get rid of that code i believe because most of that code is really incorrect anyways

bholagabbar: What exactly do you want me to do?

omod/src/main/java/org/openmrs/web/controller/ForgotPasswordFormController.java
wluyima: It is because the GP isn't set, was this in a unit test? With an arbitrary value like 8, the code would work if the GP is blnak but would break if the admin has set the GP value in production

omod/src/test/java/org/openmrs/web/controller/ForgotPasswordFormControllerTest.java
wluyima: Just un comment the line the calls changePassword because that would be the original behavior

omod/src/test/java/org/openmrs/web/controller/ForgotPasswordFormControllerTest.java
wluyima: Typically, you can look at the code and see what it does when a user provides an invalid secret question, from this [code](https://github.com/openmrs/openmrs-module-legacyui/blob/master/omod/src/main/java/org/openmrs/web/controller/ForgotPasswordFormController.java#L184), the user is shown the form again, if you're getting back null, possibly there is something wrong happening

wluyima: I could be wrong but i would assume showForm should return a non null value, can you print out the value returned by showForm when you run the application just to confirm?

omod/src/test/java/org/openmrs/web/controller/ForgotPasswordFormControllerTest.java
wluyima: Since the controller is authenticating the user, then i guess it is okay to leave the method name saying authenticate, and it's okay to call logout

PR #33, opened 2016-01-21
FINAL fix with tests for LUI-45 [Forgot Password Form Leaks Valid Usernames]

bholagabbar:

<br>
**In light of recent comments on the ticket, it has come to notice that this is a serious security issue and must be resolved ASAP. If you see the comments, you will notice that this code fixes the issue as per ticket description and quite well in general.**
<br>

**basically, the hacker will never know if the user exists because a question is asked irrespective of whether a user is present and invalid users are never authenticated and locked out after 5 tries**

### DIFF
--- a/omod/src/main/java/org/openmrs/web/attribute/handler/ConceptFieldGenDatatypeHandler.java
+++ b/omod/src/main/java/org/openmrs/web/attribute/handler/ConceptFieldGenDatatypeHandler.java
@@ -19,7 +19,7 @@ import org.springframework.stereotype.Component;
 /**
  * Handler for the Concept custom datatype
  */
-//@Component
+@Component
 public class ConceptFieldGenDatatypeHandler extends SerializingFieldGenDatatypeHandler<ConceptDatatype, Concept> {
 	
 	/**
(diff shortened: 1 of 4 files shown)


omod/src/main/java/org/openmrs/web/controller/ForgotPasswordFormController.java
```diff
@@ -143,45 +149,78 @@ protected ModelAndView onSubmit(HttpServletRequest request, HttpServletResponse
...
+		
 		request.setAttribute("uname", username);
+		
 		return showForm(request, response, errors);
 	}
 	
-}
+	public String getRandomPassword() {
+		//Password should be of length at-least 8, must have 1 upper case letter and 1 number
+		return RandomStringUtils.randomAlphabetic(1).toUpperCase() + RandomStringUtils.randomAlphanumeric(8)+ RandomStringUtils.randomNumeric(1);
```
wluyima: The generated password needs to be longer than the value of the required minimum password length, an arbitrary number like 8 should be used as the default value  in case the global property is not set otherwise this will fail if the amdin has set the global property to a longer value like 12

PR #34, opened 2016-01-21
FINAL fix addressing security issues in LUI-45 [Forgot Password Form Leaks Valid Usernames]

bholagabbar:

<br>
**In light of recent comments on the ticket, it has come to notice that this is a serious security issue and must be resolved ASAP. If you see the comments, you will notice that this code fixes the issue as per ticket description and quite well in general.**
<br>

**basically, the hacker will never know if the user exists because a question is asked irrespective of whether a user is present and invalid users are never authenticated and locked out after 5 tries**

### DIFF
--- a/omod/src/main/java/org/openmrs/web/controller/ForgotPasswordFormController.java
+++ b/omod/src/main/java/org/openmrs/web/controller/ForgotPasswordFormController.java
@@ -9,8 +9,10 @@
  */
 package org.openmrs.web.controller;
 
+import java.util.ArrayList;
 import java.util.Date;
 import java.util.HashMap;
+import java.util.List;
 import java.util.Map;
 
 import javax.servlet.ServletException;
@@ -18,9 +20,9 @@ import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
 import javax.servlet.http.HttpSession;
 
+import org.apache.commons.lang.RandomStringUtils;
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
-import org.apache.commons.lang.StringUtils;
 import org.openmrs.User;
 import org.openmrs.api.context.Context;
 import org.openmrs.util.PrivilegeConstants;
@@ -68,7 +70,7 @@ public class ForgotPasswordFormController extends SimpleFormController {
 	 *      org.springframework.validation.BindException)
 	 */
 	protected ModelAndView onSubmit(HttpServletRequest request, HttpServletResponse response, Object obj,
-	        BindException errors) throws Exception {
+	                                BindException errors) throws Exception {
 		
 		HttpSession httpSession = request.getSession();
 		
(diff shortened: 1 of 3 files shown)


omod/src/main/java/org/openmrs/web/controller/ForgotPasswordFormController.java
```diff
@@ -143,45 +149,78 @@ protected ModelAndView onSubmit(HttpServletRequest request, HttpServletResponse
...
+		
 		request.setAttribute("uname", username);
+		
 		return showForm(request, response, errors);
 	}
 	
-}
+	public String getRandomPassword() {
+		//Password should be of length at-least 8, must have 1 upper case letter and 1 number
+		return RandomStringUtils.randomAlphabetic(1).toUpperCase() + RandomStringUtils.randomAlphanumeric(8)+ RandomStringUtils.randomNumeric(1);
```
wluyima: Still forgot this? Is this the new pull request?

omod/src/main/java/org/openmrs/web/controller/ForgotPasswordFormController.java
```diff
@@ -143,45 +149,78 @@ protected ModelAndView onSubmit(HttpServletRequest request, HttpServletResponse
...
+		
+		questions.add(Context.getMessageSourceService().getMessage("What is your best friend's name?"));
+		questions.add(Context.getMessageSourceService().getMessage("What is your grandfather's home town?"));
+		questions.add(Context.getMessageSourceService().getMessage("What is your mother's maiden name?"));
+		questions.add(Context.getMessageSourceService().getMessage("What is your favorite band?"));
+		questions.add(Context.getMessageSourceService().getMessage("What is your first pet's name?"));
+		questions.add(Context.getMessageSourceService().getMessage("What is your brother's middle name?"));
+		questions.add(Context.getMessageSourceService().getMessage("Which city were you born in?"));
+		
+		int hashValueForName = username.hashCode();
```
wluyima: Apache's RandomUtils can easily give you a random number with a single method call

omod/src/main/java/org/openmrs/web/controller/ForgotPasswordFormController.java
bholagabbar: What did I forget? You agreed yourself that I keep the number at the end. The value if fetched from global constants returns null. I put that on the last PR. If you want me to change this, then tell me what exactly you want

omod/src/main/java/org/openmrs/web/controller/ForgotPasswordFormController.java
bholagabbar: No your thinking is not correct. <br>Let me explain. If the user enters an invalid secret answer, According to your method he will be assigned a new question everytime he enters answer (which is always wrong). He will come to know that there is something wrong as question changes everytime. **The question for each username has to be unique.** That is what this is doing. HashValue will remain constant and will pick same question

omod/src/main/java/org/openmrs/web/controller/ForgotPasswordFormController.java
wluyima: See the code snippet below:
```
Integer minLength = 8;
String str = Context.getAdministrationService().getGlobalProperty(OpenmrsConstants.GP_PASSWORD_MINIMUM_LENGTH);
        if(StringUtils.isNotBlank(str)){
            minLength = Integer.valueOf(str);
        }
```
And then pass minLength to randaAlphanumeric

bholagabbar: You yourself agreed that only randomAlphaNumeric will not give correct password always. It should be RandomAlphabetic(1).toUppercase() [worst case, randomAlphanumberic returns all numbers. Hence this is needed] + RandomAlphanumeric (8) + RandomNumeric(1) [worst case..all are letters]. *My code addresses this all. Why would you want me to change such a small detail ? It is almost cruel after so many PRs...hope you don't mind*

wluyima: That's not what mean, don't worry about it i will fix it as i merge the code

omod/src/main/java/org/openmrs/web/controller/ForgotPasswordFormController.java
wluyima: Fair point

PR #35, opened 2016-01-23
Adding fixes to https://issues.openmrs.org/browse/LUI-41

rasanjanap:


omod/src/main/java/org/openmrs/web/controller/patient/ShortPatientFormValidator.java
```diff
@@ -89,12 +89,18 @@ public void validate(Object obj, Errors errors) {
...
+
 			if (OpenmrsUtil.nullSafeEqualsIgnoreCase(possibleDuplicate.getGivenName(), personName.getGivenName())
-			        && OpenmrsUtil.nullSafeEqualsIgnoreCase(possibleDuplicate.getMiddleName(), personName.getMiddleName())
-			        && OpenmrsUtil.nullSafeEqualsIgnoreCase(possibleDuplicate.getFamilyName(), personName.getFamilyName())) {
-				errors.reject("Patient.duplicateName", new Object[] { personName.getFullName() }, personName.getFullName()
-				        + " is a duplicate name for the same patient");
+					&& OpenmrsUtil.nullSafeEqualsIgnoreCase(possibleDuplicate.getMiddleName(), personName.getMiddleName())
+					&& OpenmrsUtil.nullSafeEqualsIgnoreCase(possibleDuplicate.getFamilyName(), personName.getFamilyName())) {
+					if (possibleDuplicate.isVoided()) {
+						errors.reject("Patient.duplicateName", new Object[]{personName.getFullName()}, personName.getFullName()
```
wluyima: Can you actually use a message code and have this translated in the message.properties file, can you also do the same for the other error message too? Thanks!

wluyima: Never mind, i will do it as i merge the code

wluyima:
Good practice is to create a new branch for each ticket your work on instead of working off master because you could be working on multiple tickets and you don't want to mix up the changes for different tickets

rasanjanap:
Got it @wluyima thanks for the tip

omod/src/main/java/org/openmrs/web/controller/patient/ShortPatientFormValidator.java
rasanjanap: Do I need to do any more things?

wluyima: Nope, thanks!

PR #36, opened 2016-01-26
Redirecting Authenticated User to home if user tries to access 'login.htm' page [LUI-60]

bholagabbar:
Issue: [[LUI 60] login.htm Should redirect to previous/home page if logged in in legacy UI](https://issues.openmrs.org/browse/LUI-60)<br>

If the User is Authenticated, then he should be redirected to his/her OpenMRS home page and should not be allowed the access the login page

### DIFF
--- a/omod/src/main/java/org/openmrs/web/controller/LoginController.java
+++ b/omod/src/main/java/org/openmrs/web/controller/LoginController.java
@@ -43,9 +43,10 @@ public class LoginController {
 			}
 			webRequest.removeAttribute(WebConstants.INSUFFICIENT_PRIVILEGES, WebRequest.SCOPE_SESSION);
 		}
-		
-		//If there is a currently logged in user and they failed a privilege check, else go to login in page
-		if (Context.getAuthenticatedUser() != null && failedPrivilegeCheck) {
+		//If the user is logged in and there are no privilege problems, he should be redirected to home
+		if(Context.getAuthenticatedUser() != null && !failedPrivilegeCheck) {
+			return "redirect:/index.htm";
+		} else if (Context.getAuthenticatedUser() != null && failedPrivilegeCheck) { //Else, if there is a currently logged in user and they failed a privilege check, else go to login in page
 			model.addAttribute("foundMissingPrivileges", true);
 			webRequest.setAttribute(WebConstants.OPENMRS_ERROR_ATTR, "error.insufficientPrivileges",
 			    WebRequest.SCOPE_SESSION);


omod/src/main/java/org/openmrs/web/controller/LoginController.java
```diff
@@ -43,9 +43,10 @@ public String handleRequest(WebRequest webRequest, ModelMap model) {
 			}
 			webRequest.removeAttribute(WebConstants.INSUFFICIENT_PRIVILEGES, WebRequest.SCOPE_SESSION);
 		}
-		
-		//If there is a currently logged in user and they failed a privilege check, else go to login in page
-		if (Context.getAuthenticatedUser() != null && failedPrivilegeCheck) {
+		//If the user is logged in and there are no privilege problems, he should be redirected to home
+		if(Context.getAuthenticatedUser() != null && !failedPrivilegeCheck) {
+			return "/module/legacyui"+"/index";
```
djazayeri: I would expect that you would return something starting with "forward:". We don't want to just display the "index" view with an empty model, but rather to forward to that page.

k-joseph: what advantage does separating these two strings into two string objects concatenated together introduce?

bholagabbar: @k-joseph : No it was just for consistency. If you see, at the bottom of the page it is '/module/legacyui+LOGIN_FORM'

bholagabbar: djazayeri: The ticket description asked to bring the user to the home page. So I thought this would be best. What page would you suggest we redirect him to? I'm not sure I understand the "forward:" refernce

bholagabbar:
I'd just like to ask if the basic logic for redirecting the user seems fine? Because I didn't want to confuse anything with the condition where it checks for user with missing privileges. I mean I hope the flow of control is fine

omod/src/main/java/org/openmrs/web/controller/LoginController.java
k-joseph: i see no need here for such consistence @bholagabbar you are referring to

bholagabbar: @k-joseph Sure no issues I'll keep it as one string. What page do you suggest I redirect to as @djazayeri suggested?

djazayeri: Sorry, my mistake, it should be "redirect:" not forward. Basically you don't want the URL to keep saying "login.htm" but you happen to display the home screen. Rather if they try to go to the login screen, but they are already logged in, then you should _redirect_ them so that the browser loads the home page instead.

bholagabbar: @djazayeri Ahh thanks! Got it. I"ll do that :D

bholagabbar: @djazayeri I've done `"redirect:/index.htm";` and it seems to work fine. Are you fine with this?

djazayeri: fine by me

k-joseph:
you can go ahead and commit so that we can review what you've got as a whole!

PR #37, opened 2016-01-29
LUI-48:Retired Answer Concepts show HTML code in the list

vishnuraom:


vishnuraom:
@dkayiwa I have added a message code and a method which retrieve's it(according to users locale).

omod/src/main/java/org/openmrs/web/controller/ConceptFormController.java
```diff
@@ -1036,7 +1036,8 @@ public int getNumberOfObsUsingThisConcept() {
 					name = answer.getAnswerDrug().getFullName(locale);
 				}
 				if (answer.getAnswerConcept().isRetired()) {
-					name = "<span class='retired'>" + name + "</span>";
+					String string = Context.getMessageSourceService().getMessage("Concept.retired");
```
wluyima: The name for a retired concept name needs to have a strike through, so the original code needs to be tweaked a little, may be you need to use html codes for the tags e.g **`&lt;`** for  **<**

wluyima: So you won't need the label for Retired

vishnuraom:
@wluyima The strike through is visible if I remove the "c:out" tag on line 7 of https://gist.github.com/vishnuraom/119a939a6f5d0ec0a3c9#file-concept-jsp but on https://talk.openmrs.org/t/concept-units-w-superscript-dont-display-properly-in-ui-since-1-11/3219 they said that removing "c:out" tag would provide a scope for XSS attacks.
I thought that the above case is valid is here.

vishnuraom:
@dkayiwa I have to check whether a answer concept is retired before printing (Retired) after concept id(In Concept.jsp) but getConceptAnswers in ConceptFormController is  returning a Map<String, String>(not ConceptAnswer's) so I'm unable to find whether a answer concept is retired or not.

wluyima:
You need to use a strike through for retired answers instead of a text for retired

omod/src/main/webapp/dictionary/concept.jsp
```diff
@@ -227,7 +227,7 @@
                 <c:forEach items="${command.locales}" var="loc">
                     <td class="${loc}">
                         <c:forEach items="${command.conceptAnswersByLocale[loc]}" var="answer">
-                            <a href="concept.htm?conceptId=${fn:substring(answer.key, 0, fn:indexOf(answer.key, '^'))}"><c:out value="${answer.value}" /> (${fn:substring(answer.key, 0, fn:indexOf(answer.key, '^'))})</a><br/>
+                            <a href="concept.htm?conceptId=${fn:substring(answer.key, 0, fn:indexOf(answer.key, '^'))}">${answer.value}(${fn:substring(answer.key, 0, fn:indexOf(answer.key, '^'))})</a><br/>
```
wluyima: I thought you need to use <c:out to avoid xss attacks?

wluyima: Does this strike through retired answers?

wluyima: To achieve the strike through while maintaining the c:out tag just add the **retired** class to the surrounding tag

vishnuraom: @wluyima The strike off is visible if I remove the <c:out ,This can be seen at https://issues.openmrs.org/secure/attachment/43995/Strikeoff_visible.png

wluyima: But you need to use the c:out by convention, there is issues with not using it. You need to use a retired class on one of the surrounding tags

vishnuraom: @wluyima Did you mean adding a retired class like in here, https://github.com/openmrs/openmrs-module-legacyui/blob/master/omod/src/main/java/org/openmrs/web/controller/ConceptFormController.java#L1046 
Even if the retired class is added, along with <c:out the HTML code is being displayed.

wluyima: That's what we need to fix that while still using c:out tag, the correct thing happens, i believe there is a way

vishnuraom: @wluyima I'll try to find a work around and please let me know if you have any ideas regarding the issue.

wluyima: Cool

dkayiwa:
@vishnuraom did you get a work around?

vishnuraom:
@dkayiwa Sorry,I was not able to find any work around.There is also a ticket https://issues.openmrs.org/browse/TRUNK-4762 where the issue is with &lt;c:out&gt; tag.But the solution proposed there can't be used here as the strikeoff should be visible.

omod/src/main/webapp/dictionary/concept.jsp
wluyima: You need to use c:out tag otherwise you'll get other bugs

dkayiwa:
Closing because this was rejected.

PR #38, opened 2016-02-04
[LUI-61] Add support for back button on patient search page

jennparise:


omod/src/main/webapp/resources/scripts/jquery-ui/js/openmrsSearch.js
```diff
@@ -546,26 +546,38 @@ function OpenmrsSearch(div, showIncludeVoided, searchHandler, selectionHandler,
...
+        		if (inputNodeDOM.setSelectionRange) {
+        			inputNodeDOM.setSelectionRange(lastSearchTextLength,lastSearchTextLength);
+        		}
+        	}
+       
+                var keyEvent = jQuery.Event("keyup");    
+                //jQuery normalizes event.which fixing some issues with not being able to trigger enter key 
+                keyEvent.which = 13;
                 $j("#inputNode").trigger(keyEvent);
             }
```
wluyima: Seems to me like we are doing more than what needs to be done to achieve this, the widgets already have a way to support triggering a search on page load given an initial search phrase, it configured as searchPhrase in the widget options, you should just read the last search text from the browser history and set it as the value of the searchPhrase widget option and things will happen magically

jennparise: Thanks for the feedback, I will fix it to use the searchPhrase then.

wluyima: We might want to do something similar for includeVoided and showVerbose, i think there is a commit that added an option for storing these

jennparise: I saw that, I am in the process of adding it now. I am assuming we are still using the histoy.pushState and adding the parameters there and then reading them from the url? I may be getting off because of the code that was already written but if we are going to put back button functionality in openmrsSearch then we don't use the lastSearchParams, right? Just curious too, I have used sessionStorage in the past what is the benefit of using the history.pushState?

omod/src/main/webapp/resources/scripts/jquery-ui/js/openmrsSearch.js
```diff
@@ -546,26 +546,38 @@ function OpenmrsSearch(div, showIncludeVoided, searchHandler, selectionHandler,
             });
 
             // Browser back button support, for preserve data
-            if (lastSearchParams !== null) {
-                $j('#inputNode').val(lastSearchParams.lastSearchText);
-                if (lastSearchParams.includeVoided == 1) {
+            var lastSearchText = decodeURIComponent((new RegExp('[?|&]' + 'lastSearchText'
```
wluyima: I would assume you don't need to do this, lastSearchText, includeVoided and includeVerbose should be stored in the state object you pass to pushState, that way you don't have to do all this but instead read them from the history's state

omod/src/main/webapp/resources/scripts/jquery-ui/js/openmrsSearch.js
```diff
@@ -546,26 +546,38 @@ function OpenmrsSearch(div, showIncludeVoided, searchHandler, selectionHandler,
             });
 
             // Browser back button support, for preserve data
-            if (lastSearchParams !== null) {
-                $j('#inputNode').val(lastSearchParams.lastSearchText);
-                if (lastSearchParams.includeVoided == 1) {
+            var lastSearchText = decodeURIComponent((new RegExp('[?|&]' + 'lastSearchText' 
+            		+ '=' + '([^&;]+?)(&|#|;|$)').exec(location.search)||[,""])[1].replace(/\+/g, '%20')) || null;
+            var includeVoided = decodeURIComponent((new RegExp('[?|&]' + 'includeVoided'
```
wluyima: includeVoided and includeVerbose in fact should be handled exactly the way we are doing the initial search text

omod/src/main/webapp/resources/scripts/jquery-ui/js/openmrsSearch.js
```diff
@@ -546,26 +546,38 @@ function OpenmrsSearch(div, showIncludeVoided, searchHandler, selectionHandler,
...
+            		+ '=' + '([^&;]+?)(&|#|;|$)').exec(location.search)||[,""])[1].replace(/\+/g, '%20')) || null;
+            var includeVoided = decodeURIComponent((new RegExp('[?|&]' + 'includeVoided' 
+            		+ '=' + '([^&;]+?)(&|#|;|$)').exec(location.search)||[,""])[1].replace(/\+/g, '%20')) || null;
+            var includeVerbose = decodeURIComponent((new RegExp('[?|&]' + 'includeVerbose' 
+            		+ '=' + '([^&;]+?)(&|#|;|$)').exec(location.search)||[,""])[1].replace(/\+/g, '%20')) || null;
+            
+            //if we have lastSearchText then set input text, move cursor to the end and trigger a keyup event to load the search
+            if (lastSearchText !== null) {
+            	
+                $j('#inputNode').val(lastSearchText);
```
wluyima: Instead of this, you need to assign the value of lastSearchText to options.searchPhrase instead, you will need to make the widget support showVerbose and includeVoided as widget options too.

omod/src/main/webapp/resources/scripts/jquery-ui/js/openmrsSearch.js
```diff
@@ -951,6 +963,13 @@ function OpenmrsSearch(div, showIncludeVoided, searchHandler, selectionHandler,
 
         _doSelected: function(position, rowData) {
             if(this.options.selectionHandler) {
+		// Check the browser compatibility and, add an entry to history
+		if (window.history.pushState) {
+			window.history.pushState(
+				{}, "",
+				document.location.pathname + "?lastSearchText=" + document.getElementById('inputNode').value
```
wluyima: You should not set lastSearchText as a request parameter, it should be in the state object you push into the history, same for includeVoided and showVerbose

wluyima: Only JSPs should use request parameters

omod/src/main/webapp/resources/scripts/jquery-ui/js/openmrsSearch.js
jennparise: Ah, I haven't used history.pushState. I see how to do it now! You can learn so much from even such a little fix. Thanks for your help.

wluyima: It is being used below towards the bottom of this pull request

wluyima: There are 2 ways you can read the state values from the history, you can provide onpopstate which is a call back function or you can directly read the state from the history object using history.state

wluyima: Not sure which one is better though

jennparise:
I am making a mess out of Github. I was trying to squash and now it seems I went back too far. How do I get get this down to my own single commit? I have been trying to figure it out on my own but I am just not sure how to do it with the pull request and not make a mess.

jennparise:
Ok, I hope I did not bombard anyone with too many updates. I think I finally have the hang of dealing with changes while I have a pull request. The push -f option and squashing scared me but I think I know how to use it now. I believe I am ready for feedback on my latest commit. Thanks for your help so far. I really am interested in learing how all this works so I appreciate all of your explanations.

omod/src/main/webapp/resources/scripts/jquery-ui/js/openmrsSearch.js
```diff
@@ -545,30 +547,23 @@ function OpenmrsSearch(div, showIncludeVoided, searchHandler, selectionHandler,
...
-                if (lastSearchParams.includeVoided == 1) {
-                    $j('#includeVoided').attr('checked','checked');
-                }
-                if (lastSearchParams.includeVerbose == 1) {
-                    $j('#includeVerbose').attr('checked','checked');
+            //Browser back button support, to preserve data
+            if(history.state || lastSearchParams.lastSearchText){
+            	//check for lastSearchParams so we don't overwrite something sent as a option
+                //if it is not set check to see if there are parameters to get from history.pushState
+            	if(!lastSearchParams.lastSearchText){
```
wluyima: i would assume that searchPhrase, includeVoided and showVerbose should all be treated just the same, priority should be the widget options then history. You can also copy the history into the options just like the search phrase and get rid of lastSearchParams

wluyima: Correction,  priority should be first the history and then widget options

omod/src/main/webapp/resources/scripts/jquery-ui/js/openmrsSearch.js
```diff
@@ -586,11 +581,30 @@ function OpenmrsSearch(div, showIncludeVoided, searchHandler, selectionHandler,
                 else
                     self._table.numberOfPages = Math.floor(initialRowCount/self._table.fnSettings()._iDisplayLength)+1;
 
-            } else if(self.options.searchPhrase || self.options.doSearchWhenEmpty) {
-                if (self.options.searchPhrase == null) {
-                    self.options.searchPhrase = "";
+            } else{
```
wluyima: Why is this changing from else if to else?

omod/src/main/webapp/resources/scripts/jquery-ui/js/openmrsSearch.js
wluyima: You can ignore my comments above

omod/src/main/webapp/resources/scripts/jquery-ui/js/openmrsSearch.js
```diff
@@ -586,11 +581,30 @@ function OpenmrsSearch(div, showIncludeVoided, searchHandler, selectionHandler,
...
+                		$j('#includeVoided').attr('checked','checked');
+                	}
+                	$j(input).val(self.options.searchPhrase).keyup();
+                	//setting the cursor to the end of the searchPhrase
+                    //setSelectionRange is a method of the DOM element not the jQuery object so get the DOM element
+                    var inputNodeDOM = $j('#inputNode').get(0);
+                    if(inputNodeDOM){
+                    	var searchPhraseLength = self.options.searchPhrase.length;
+                    	if (inputNodeDOM.setSelectionRange) {
+                    		inputNodeDOM.setSelectionRange(searchPhraseLength, searchPhraseLength);
```
wluyima: Some one added this as part of the 2 commits someone made, is it really needed?

omod/src/main/webapp/resources/scripts/jquery-ui/js/openmrsSearch.js
jennparise: Just to clarify, do I need to make changes here?

omod/src/main/webapp/resources/scripts/jquery-ui/js/openmrsSearch.js
jennparise: This is where the search gets triggered when the search box gets populated by passed in data rather than the user typing in data to trigger a keyup. Is there something I am not seeing that should make that happen somewhere else?

omod/src/main/webapp/resources/scripts/jquery-ui/js/openmrsSearch.js
jennparise: That was an oversight when I redid that section I have added a new commit to fix this.

omod/src/main/webapp/resources/scripts/jquery-ui/js/openmrsSearch.js
wluyima: No

omod/src/main/webapp/resources/scripts/jquery-ui/js/openmrsSearch.js
```diff
@@ -587,10 +582,27 @@ function OpenmrsSearch(div, showIncludeVoided, searchHandler, selectionHandler,
...
+            		$j('#includeVoided').attr('checked','checked');
+            	}
+            	$j(input).val(self.options.searchPhrase).keyup();
+            	//setting the cursor to the end of the searchPhrase
+            	//setSelectionRange is a method of the DOM element not the jQuery object so get the DOM element
+            	var inputNodeDOM = $j('#inputNode').get(0);
+            	if(inputNodeDOM){
+            		var searchPhraseLength = self.options.searchPhrase.length;
+            		if (inputNodeDOM.setSelectionRange) {
+            			inputNodeDOM.setSelectionRange(searchPhraseLength, searchPhraseLength);
```
wluyima: This is not what triggers the search, line 596 is what triggers it if the user doesn't provide any input and before it was outside this else-if clause, just curious if there is a reason why you moved it into the else-if statement?

jennparise: I guess I was misunderstanding your previous comment. What are you not sure we need from previous commits? I thought you were refering to the whole section here. I redid this section trying to make sure we had a search loaded when the checkboxes were checked and I guess in my logic I was only doing the keyup when we had a searchPhrase. I can move the keyup and cursor code outside of the else if. I should have double checked the original functionality and realized it was meant to work in other scenarious too because it will still trigger it when I need it for the checkboxes too.

wluyima: Yes, this code lines form 599 were added by one of the earlier commits and i don't find them really necessary so it would be nice if you removed them, and yes the call to keyup() is supposed to be inside the else-if clause, thanks in advance! [This is](https://github.com/openmrs/openmrs-module-legacyui/blob/master/omod/src/main/webapp/resources/scripts/jquery-ui/js/openmrsSearch.js#L589-L594) how it was before

wluyima: You can keep the keypup() call inside the else-if since it was the case before

jennparise: Sorry, 599:var inputNodeDOM = $j('#inputNode').get(0); I added to get the input box for setting the selection range and moving the cursor. Is that what you are refering too? Am I doing too many checks?

wluyima: I guess am trying to understand why that is necessary, i.e setting selection range and setting the cursor, because this feature worked before without it

jennparise: Oh, I had seen in the comments that the cursor should go to the end of the text rather than at the beginning. Is this not supposed to be done?

wluyima: What the logic seems to be doing is to select(highlight) the text set in the search input box although the comment says it's placing the cursor to the end, i personally see no benefit to it and i suspect it can fail in some browsers. Can you test it out in the browser to see the behavior?

jennparise: This code is what sets the cursor to the end of the text for firefox and for chrome. I was not able to find a way to test on IE. I have a check though that if selectionRange is not browser supported it will ignore it which is what should happen for IE. There is createRange for IE i believe but I did not want to add code I could not see the behavior for since I could not track down a copy of IE ( I only run Ubuntu). I can verify that it works with Firefox and Chrome however. Would you rather I remove it still? If you would like for it to stay I do need to add a new commit. I accidentally lost the code when I commited for getting the focus for the chrome browser.

jennparise: Also, if you know of a better way to do this, please let me know. This is just how I have done it in the past. If you know of something else I am always happy to learn new ways.

wluyima: If it works, that's fine

omod/src/main/webapp/resources/scripts/jquery-ui/js/openmrsSearch.js
```diff
@@ -104,6 +104,8 @@ function OpenmrsSearch(div, showIncludeVoided, searchHandler, selectionHandler,
  *   verboseHandler: function to be called to return the text to display as verbose output
  *   attributes: Array of names for attributes types to display in the list of results
  *   showSearchButton: Boolean, indicating whether to use search button for immediate search
+ *   showVerbose: Boolean, indicates whether to check checkbox for verbose 
+ *   includeVoided: Boolean, indicates whether to check checkbox for voided 
  *   lastSearchParams: Object with properties lastSearchText, includeVoided and includeVerbose, to preserve data with browser back button
```
wluyima: You should not use request parameters, so you need to get rid of this and update the search jsp pages using them not to set them. See my comment on the ticket

omod/src/main/webapp/resources/scripts/jquery-ui/js/openmrsSearch.js
```diff
@@ -545,30 +545,24 @@ function OpenmrsSearch(div, showIncludeVoided, searchHandler, selectionHandler,
...
-                if (el.attr('id') === "findPatients") {
-                    var text = $j('#inputNode').val();
-                    codex = $j('#inputNode').val().charCodeAt($j('#inputNode').val().length - 1);
+            //Browser back button support, to preserve data
+            if(typeof(Storage) !== "undefined") {
+            	//check to see if there are parameters to get from sessionStorage
+            	//after setting the fields clear out session storage so it doesn't get used on other pages
+            	if(sessionStorage.lastSearchText){
+            		self.options.searchPhrase = sessionStorage.lastSearchText;  
+            		sessionStorage.lastSearchText = "";
```
wluyima: You need to remove it instead with the removeItem method,below is how you do it:

`sessionStorage.removeItem("lastSearchText");`

omod/src/main/webapp/resources/scripts/jquery-ui/js/openmrsSearch.js
```diff
@@ -951,6 +963,12 @@ function OpenmrsSearch(div, showIncludeVoided, searchHandler, selectionHandler,
 
         _doSelected: function(position, rowData) {
             if(this.options.selectionHandler) {
+        		// Check the browser compatibility and, add an entries to sessionStorage
+            	if(typeof(Storage) !== "undefined") {
+        	        sessionStorage.lastSearchText= $j('#inputNode').val();
+        			sessionStorage.includeVerbose = ($j(verboseCheckBox).is(':checked')) ? "checked" : "unchecked";
```
wluyima: Why not just store true/false since these are boolean values

omod/src/main/webapp/resources/scripts/jquery-ui/js/openmrsSearch.js
```diff
@@ -951,6 +963,12 @@ function OpenmrsSearch(div, showIncludeVoided, searchHandler, selectionHandler,
...
+        			sessionStorage.includeVoided = ($j(checkBox).is(':checked')) ? "checked" : "unchecked";
```
wluyima: Same for this

omod/src/main/webapp/resources/scripts/jquery-ui/js/openmrsSearch.js
```diff
@@ -587,11 +581,29 @@ function OpenmrsSearch(div, showIncludeVoided, searchHandler, selectionHandler,
...
             }
+            $j(input).val(self.options.searchPhrase).keyup();
+        	//setting the cursor to the end of the searchPhrase
+        	//setSelectionRange is a method of the DOM element not the jQuery object so get the DOM element
+        	var inputNodeDOM = $j('#inputNode').get(0);
+        	inputNodeDOM.focus();
+        	if(inputNodeDOM){
+        		var searchPhraseLength = self.options.searchPhrase.length;
+        		if (inputNodeDOM.setSelectionRange) {
+        			inputNodeDOM.setSelectionRange(searchPhraseLength, searchPhraseLength);
```
wluyima: What is the purpose of setting the selection range?

omod/src/main/webapp/resources/scripts/jquery-ui/js/openmrsSearch.js
wluyima: You could possibly store any random text you want, preferably 'true' but i don't you need 2 of them for each state of the checkbox, presence of the single text you choose to use should be enough, if it doesn't exist then the checkbox wasn't checked

omod/src/main/webapp/resources/scripts/jquery-ui/js/openmrsSearch.js
```diff
@@ -587,11 +581,30 @@ function OpenmrsSearch(div, showIncludeVoided, searchHandler, selectionHandler,
...
+            	}           	
             }
+            $j(input).val(self.options.searchPhrase).keyup();
+        	//setting the cursor to the end of the searchPhrase
+        	//setSelectionRange is a method of the DOM element not the jQuery object so get the DOM element
+        	var inputNodeDOM = $j('#inputNode').get(0);
+        	inputNodeDOM.focus();
+        	if(inputNodeDOM){
+        		var searchPhraseLength = self.options.searchPhrase.length;
+        		//setSelectionRange is for setting the cursor at the end of the search text
```
wluyima: I thought setSelectionRange sets a range of text to highlight [see this](https://developer.mozilla.org/en-US/docs/Web/API/HTMLInputElement/setSelectionRange), by the way which doesn't work on my machine and I'm running firefox on OSX, putting the cursor to the end of the text is something else.

jennparise: This is actually a common way of putting the cursor at the end of text. It works for Chrome and Firefox on my Ubuntu machine. If you are not satisfied with this solution though I am happy to either not set the cursor at the end or to try something else.

PR #41, opened 2016-02-16
TRUNK-4734/LUI-65 Implement a new CheckBox which when selected creates a Provider Account for the user being created in user.form

bholagabbar:
Fix for [TRUNK-4734/LUI-65](https://issues.openmrs.org/browse/LUI-65): Implement a new CheckBox which when selected creates a Provider Account for the user being created in user.form

[Image1](https://issues.openmrs.org/secure/attachment/44399/final.jpg)
[Image2](https://issues.openmrs.org/secure/attachment/44398/final2.jpg)

### DIFF
--- a/api/src/main/resources/messages.properties
+++ b/api/src/main/resources/messages.properties
@@ -1,3 +1,8 @@
 ${project.parent.artifactId}.title=Legacy UI Module
 ${project.parent.artifactId}.manage=Manage module
 ${project.parent.artifactId}.patient.duplicateName.voided={0} is a duplicate name for the same patient and it's voided. Please restore the existing name instead of creating new name
+
+legacyui.manageuser.providerAccount=Provider Account
+legacyui.manageuser.createProviderAccount=Create a Provider account for this user
+legacyui.manageuser.providerIdentfier=Provider Identifier(s):
+legacyui.manageuser.noProviderIdentifier=No Identifier Specified
--- a/omod/src/main/java/org/openmrs/web/controller/user/UserFormController.java
+++ b/omod/src/main/java/org/openmrs/web/controller/user/UserFormController.java
@@ -20,6 +20,7 @@ import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
 import org.openmrs.Person;
 import org.openmrs.PersonName;
+import org.openmrs.Provider;
 import org.openmrs.Role;
 import org.openmrs.User;
 import org.openmrs.api.PasswordException;
@@ -72,8 +73,7 @@ public class UserFormController {
 		User u = null;
 		try {
 			u = Context.getUserService().getUser(Integer.valueOf(userId));
-		}
-		catch (Exception ex) {
+		} catch (Exception ex) {
 			log.error("Error while getting user", ex);
 		}
 		if (u == null) {
@@ -104,9 +104,10 @@ public class UserFormController {
 	}
 	
 	@RequestMapping(value = "/admin/users/user.form", method = RequestMethod.GET)
-	public String showForm(@RequestParam(required = false, value = "userId") Integer userId,
-	        @RequestParam(required = false, value = "createNewPerson") String createNewPerson,
-	        @ModelAttribute("user") User user, ModelMap model) {
+	public String showForm(
+			@RequestParam(required = false, value = "userId") Integer userId, 
+			@RequestParam(required = false, value = "createNewPerson") String createNewPerson, 
+			@ModelAttribute("user") User user, ModelMap model) {
 		
 		// the formBackingObject method above sets up user, depending on userId and personId parameters   
 		
@@ -123,6 +124,13 @@ public class UserFormController {
 			model.addAttribute("changePassword", new UserProperties(user.getUserProperties()).isSupposedToChangePassword());
 			
 			model.addAttribute("secretQuestion", userService.getSecretQuestion(user));
+		}
+		
+		if (user.getPerson().getId() != null && !Context.getProviderService().getProvidersByPerson(user.getPerson()).isEmpty()) {
+			model.addAttribute("isProvider", true);
+			model.addAttribute("providerList", Context.getProviderService().getProvidersByPerson(user.getPerson()));
+		} else {
+			model.addAttribute("isProvider", false);
 		}
 		
 		// not using the default view name because I'm converting from an existing form
@@ -142,7 +150,8 @@ public class UserFormController {
 	        @RequestParam(required = false, value = "confirm") String confirm,
 	        @RequestParam(required = false, value = "forcePassword") Boolean forcePassword,
 	        @RequestParam(required = false, value = "roleStrings") String[] roles,
-	        @RequestParam(required = false, value = "createNewPerson") String createNewPerson,
+	        @RequestParam(required = false, value = "createNewPerson") String createNewPerson, 
+	        @RequestParam(required = false, value = "providerCheckBox") String addToProviderTableOption,
 	        @ModelAttribute("user") User user, BindingResult errors) {
 		
 		UserService us = Context.getUserService();
@@ -161,8 +170,7 @@ public class UserFormController {
 				Context.getUserService().purgeUser(user);
 				httpSession.setAttribute(WebConstants.OPENMRS_MSG_ATTR, "User.delete.success");
 				return "redirect:users.list";
-			}
-			catch (Exception ex) {
+			} catch (Exception ex) {
 				httpSession.setAttribute(WebConstants.OPENMRS_ERROR_ATTR, "User.delete.failure");
 				log.error("Failed to delete user", ex);
 				return "redirect:/admin/users/user.form?userId=" + request.getParameter("userId");
@@ -182,7 +190,6 @@ public class UserFormController {
 			us.unretireUser(user);
 			httpSession.setAttribute(WebConstants.OPENMRS_MSG_ATTR, "User.unRetiredMessage");
 		} else {
-			
 			// check if username is already in the database
 			if (us.hasDuplicateUsername(user)) {
 				errors.rejectValue("username", "error.username.taken");
@@ -208,8 +215,7 @@ public class UserFormController {
 			if (password.length() > 0) {
 				try {
 					OpenmrsUtil.validatePassword(user.getUsername(), password, user.getSystemId());
-				}
-				catch (PasswordException e) {
+				} catch (PasswordException e) {
 					errors.reject(e.getMessage());
 				}
 			}
@@ -284,6 +290,13 @@ public class UserFormController {
 				us.changeQuestionAnswer(user, secretQuestion, secretAnswer);
 			}
 			
+			//Check if admin wants the person associated with the user to be added to the Provider Table
+			if(addToProviderTableOption != null) {
+				Provider provider = new Provider();
+				provider.setPerson(user.getPerson());
+				provider.setIdentifier(user.getSystemId());
+				Context.getProviderService().saveProvider(provider);
+			}
 			httpSession.setAttribute(WebConstants.OPENMRS_MSG_ATTR, "User.saved");
 		}
 		return "redirect:users.list";
--- a/omod/src/main/webapp/admin/users/userForm.jsp
+++ b/omod/src/main/webapp/admin/users/userForm.jsp
@@ -103,8 +103,30 @@
 			</tr>
 		</table>
 	</fieldset>	
-	
 	<br/>
+	
+    <fieldset>
+        <legend><openmrs:message code="legacyui.manageuser.providerAccount"/></legend>
+        <c:if test="${isProvider}">
+            <openmrs:message code="legacyui.manageuser.providerIdentfier"/>
+            <c:forEach items="${providerList}" var="currentProvider" varStatus="loop">   
+                <c:if test="${empty currentProvider.identifier}">
+                    <openmrs:message code="legacyui.manageuser.noProviderIdentifier"/>
+                </c:if>
+                <c:if test="${not empty currentProvider.identifier}">
+                    <c:out value="${currentProvider.identifier}"/>
+                </c:if>
+                
+                <c:if test="${!loop.last}">,</c:if>
+            </c:forEach>
+        </c:if>
+        
+        <c:if test="${!isProvider}">
+            <input id="providerCheckBox" name="providerCheckBox" type="checkbox" value="addToProviderTable"/>
+            <label for="providerCheckBox"><openmrs:message code="legacyui.manageuser.createProviderAccount"/></label>
+        </c:if>
+    </fieldset>	
+    <br/>	
 	
 	<fieldset>
 		<legend><openmrs:message code="User.loginInfo"/></legend>
--- a/omod/src/test/java/org/openmrs/web/controller/user/UserFormControllerTest.java
+++ b/omod/src/test/java/org/openmrs/web/controller/user/UserFormControllerTest.java
@@ -9,9 +9,11 @@
  */
 package org.openmrs.web.controller.user;
 
+import org.junit.Assert;
 import org.junit.Test;
 import org.openmrs.PersonName;
 import org.openmrs.User;
+import org.openmrs.api.context.Context;
 import org.openmrs.test.Verifies;
 import org.openmrs.web.test.BaseModuleWebContextSensitiveTest;
 import org.springframework.beans.factory.annotation.Autowired;
@@ -22,27 +24,50 @@ import org.springframework.validation.BindException;
 import org.springframework.web.context.request.ServletWebRequest;
 import org.springframework.web.context.request.WebRequest;
 
+
 /**
  * Tests the {@link oldUserFormController} class.
  */
 public class UserFormControllerTest extends BaseModuleWebContextSensitiveTest {
 	
+	protected static final String TEST_DATA = "org/openmrs/web/controller/include/UserFormControllerTest.xml";
+	
 	@Autowired
 	private UserFormController controller;
 	
 	/**
-	 * @see UserFormController#handleSubmission(WebRequest,HttpSession,String,String,String,null,User,BindingResult)
-	 * 
+	 * @see UserFormController#handleSubmission(WebRequest,HttpSession,String,String,String,null, String, User,BindingResult)
+	 *
 	 */
 	@Test
-	@Verifies(value = "should work for an example", method = "handleSubmission(WebRequest,HttpSession,String,String,String,null,User,BindingResult)")
+	@Verifies(value = "should work for an example", method = "handleSubmission(WebRequest,HttpSession,String,String,String,null, String, User,BindingResult)")
 	public void handleSubmission_shouldWorkForAnExample() throws Exception {
 		WebRequest request = new ServletWebRequest(new MockHttpServletRequest());
 		User user = controller.formBackingObject(request, null);
 		user.addName(new PersonName("This", "is", "Test"));
 		user.getPerson().setGender("F");
 		controller.handleSubmission(request, new MockHttpSession(), new ModelMap(), "", "Save User", "pass123", "pass123", null,
-		    null, null, new String[0], "true", user, new BindException(user, "user"));
+				null, null, new String[0], "true", null, user, new BindException(user, "user"));
+	}
+	
+	/**
+	 * @see UserFormController#handleSubmission(WebRequest,HttpSession,String,String,String,null, String, User,BindingResult)
+	 *
+	 */
+	@Test
+	@Verifies(value = "Creates Provider Account when Provider Account Checkbox is selected", method = "handleSubmission(WebRequest,HttpSession,String,String,String,null, String, User,BindingResult)")
+	public void handleSubmission_createUserProviderAccountWhenProviderAccountCheckboxIsSelected() throws Exception {
+		executeDataSet(TEST_DATA);
+		WebRequest request = new ServletWebRequest(new MockHttpServletRequest());
+		//A user with userId=2 is preset in the Test DataSet and the relevant details are passed
+		User user = Context.getUserService().getUser(2);
+		Assert.assertTrue(Context.getProviderService().getProvidersByPerson(user.getPerson()).isEmpty());
+		ModelMap model = new ModelMap();
+		model.addAttribute("isProvider", false);
+		controller.showForm(2, "true", user, model);
+		controller.handleSubmission(request, new MockHttpSession(), new ModelMap(), "", null, "Test1234", "valid secret question", "valid secret answer", "Test1234",
+				false, new String[] {"Provider"}, "true", "addToProviderTable", user, new BindException(user, "user"));
+		Assert.assertFalse(Context.getProviderService().getProvidersByPerson(user.getPerson()).isEmpty());
 	}
 	
 }
--- a/omod/src/test/resources/org/openmrs/web/controller/include/UserFormControllerTest.xml
+++ b/omod/src/test/resources/org/openmrs/web/controller/include/UserFormControllerTest.xml
@@ -1,0 +1,16 @@
+<?xml version='1.0' encoding='UTF-8'?>
+<!--
+
+    This Source Code Form is subject to the terms of the Mozilla Public License,
+    v. 2.0. If a copy of the MPL was not distributed with this file, You can
+    obtain one at http://mozilla.org/MPL/2.0/. OpenMRS is also distributed under
+    the terms of the Healthcare Disclaimer located at http://openmrs.org/license.
+
+    Copyright (C) OpenMRS Inc. OpenMRS is a registered trademark and the OpenMRS
+    graphic logo is a trademark of OpenMRS Inc.
+
+-->
+<dataset>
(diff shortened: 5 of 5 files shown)


omod/src/main/webapp/admin/users/userForm.jsp
```diff
@@ -42,6 +42,14 @@
 					</form>
 				</td>
 			</tr>
+			<tr valign="top">
+			    <td style="border-right: 1px lightgrey solid; padding-right: 5em">
+                    <h5>Add to provider table if the Role 'Provider' is selected</h5>
+                    <form method="get" action="user.form">
+                        <input id="whatDoITypeHere?" type="checkbox" value="Add Provider to Provider Table"/>
```
bholagabbar: How do I link this to code. I mean how do I maintain a boolean variable which tells me that this checkbox has been selected or not?

omod/src/main/webapp/admin/users/userForm.jsp
```diff
@@ -24,25 +24,28 @@
...
-					<h3><openmrs:message code="User.useExisting"/></h3>
-					<form method="get" action="user.form">
-						<openmrs:message code="User.whichPerson"/> <openmrs_tag:personField formFieldName="person_id" formFieldId="existingPersonId" callback="personSelectedCallback"/>
-						<br/>
-						<input id="useExistingButton" disabled="true" type="submit" value="<openmrs:message code="general.next"/>"/>
-					</form>
-				</td>
-			</tr>
-		</table>
+		<form method="get" action="user.form">
```
bholagabbar: @rkorytkowski Created one form as you asked and the data in the request is managed quite well :)

omod/src/main/java/org/openmrs/web/controller/user/UserFormController.java
```diff
@@ -53,6 +54,7 @@
 public class UserFormController {
 	
 	protected static final Log log = LogFactory.getLog(UserFormController.class);
+	protected Boolean addToProviderTable; //Need alternative rather than declaring global
```
bholagabbar: @rkorytkowski I could find no workaround other than using a global variable. Is it ok/ can you suggest something else?

omod/src/main/java/org/openmrs/web/controller/user/UserFormController.java
```diff
@@ -119,6 +123,12 @@ public String showForm(@RequestParam(required = false, value = "userId") Integer
 			model.addAttribute("createNewPerson", createNewPerson);
 		}
 		
+		if(providerTableOption != null) {
```
bholagabbar: sets the global variable value

omod/src/main/java/org/openmrs/web/controller/user/UserFormController.java
```diff
@@ -284,6 +294,21 @@ public String handleSubmission(WebRequest request, HttpSession httpSession, Mode
 				us.changeQuestionAnswer(user, secretQuestion, secretAnswer);
 			}
 			
+			//Check if the admin wants to add to Provider Option
+			if(addToProviderTable == true) {
```
bholagabbar: Code to add provider if conditions are met

omod/src/main/java/org/openmrs/web/controller/user/UserFormController.java
```diff
@@ -124,6 +125,14 @@ public String showForm(@RequestParam(required = false, value = "userId") Integer
 			
 			model.addAttribute("secretQuestion", userService.getSecretQuestion(user));
 		}
+		Collection<Provider> col = Context.getProviderService().getProvidersByPerson(user.getPerson());
```
wluyima: Get rid of this

omod/src/main/java/org/openmrs/web/controller/user/UserFormController.java
```diff
@@ -124,6 +125,14 @@ public String showForm(@RequestParam(required = false, value = "userId") Integer
...
+		
+		System.out.println(Context.getProviderService().getProvidersByPerson(user.getPerson()).isEmpty());
```
wluyima: And get rid of this too

omod/src/main/java/org/openmrs/web/controller/user/UserFormController.java
```diff
@@ -125,6 +127,13 @@ public String showForm(@RequestParam(required = false, value = "userId") Integer
 			model.addAttribute("secretQuestion", userService.getSecretQuestion(user));
 		}
 		
+		if (user.getPerson().getId() != null && !Context.getProviderService().getProvidersByPerson(user.getPerson()).isEmpty()) {
+			model.addAttribute("isProvider", true);
+			model.addAttribute("providerList", new ArrayList<Provider>(Context.getProviderService().getProvidersByPerson(user.getPerson())));
```
wluyima: Why do you need the new ArrayList here?

bholagabbar: You are right. It is not needed. I have removed it. I will commit the latest code

omod/src/main/java/org/openmrs/web/controller/user/UserFormController.java
```diff
@@ -142,7 +151,8 @@ public String handleSubmission(WebRequest request, HttpSession httpSession, Mode
 	        @RequestParam(required = false, value = "confirm") String confirm,
 	        @RequestParam(required = false, value = "forcePassword") Boolean forcePassword,
 	        @RequestParam(required = false, value = "roleStrings") String[] roles,
-	        @RequestParam(required = false, value = "createNewPerson") String createNewPerson,
+	        @RequestParam(required = false, value = "createNewPerson") String createNewPerson, 
+	        @RequestParam(required = false, value = "providerCheckBox") String addToProviderTableOption,
```
wluyima: Shouldn't this be a boolean?

bholagabbar: Naah. Not really. If the checbox is ticked, it stores the value attribute of the check box which is "addToProviderTable" else it is null

omod/src/main/java/org/openmrs/web/controller/user/UserFormController.java
```diff
@@ -284,6 +291,19 @@ public String handleSubmission(WebRequest request, HttpSession httpSession, Mode
 				us.changeQuestionAnswer(user, secretQuestion, secretAnswer);
 			}
 			
+			//Check if admin wants the person associated with the user to be added to the Provider Table
+			if(addToProviderTableOption != null) {
+				Set<Role> userRoles = user.getRoles();
+				for (Role role : userRoles) {
+					if (role.getRole().equals("Provider")) { //If 'Provider' role is selected, we add the User to the Provider Table
```
wluyima: Does the ticket say this check is needed? I don't think Provider is a standard Role for this code to be coded against a hard coded name

omod/src/main/java/org/openmrs/web/controller/user/UserFormController.java
```diff
@@ -284,6 +291,19 @@ public String handleSubmission(WebRequest request, HttpSession httpSession, Mode
...
+						Provider provider = new Provider();
+						provider.setPerson(user.getPerson());
+						provider.setName(user.getName());
```
wluyima: If you have set person, you don't need to set name too, provider.getName will take care of using the name of the linked person record

omod/src/main/java/org/openmrs/web/controller/user/UserFormController.java
```diff
@@ -284,6 +290,19 @@ public String handleSubmission(WebRequest request, HttpSession httpSession, Mode
...
+						provider.setIdentifier(user.getSystemId());
```
wluyima: If identifier is not required, i would leave it blank

omod/src/main/java/org/openmrs/web/controller/user/UserFormController.java
bholagabbar: The logic is that the user will be added to provider table iff the Provider Role is selected

omod/src/main/java/org/openmrs/web/controller/user/UserFormController.java
```diff
@@ -284,6 +290,19 @@ public String handleSubmission(WebRequest request, HttpSession httpSession, Mode
...
+						Context.getProviderService().saveProvider(provider);
```
wluyima: Is this happening after or before you save the user record? Because it should be after

omod/src/main/webapp/admin/users/userForm.jsp
```diff
@@ -103,9 +103,42 @@
 			</tr>
 		</table>
 	</fieldset>	
-	
 	<br/>
 	
+    <fieldset>
+        <legend>Provider Account</legend>
```
wluyima: Can you translate this please?

omod/src/main/java/org/openmrs/web/controller/user/UserFormController.java
bholagabbar: Ticket mentions it is needed

omod/src/main/webapp/admin/users/userForm.jsp
```diff
@@ -103,9 +103,42 @@
...
+        <table>
+            <tr>
+                <td>
+                    <c:if test="${isProvider == true}">
+                        <table cellspacing="15">
+                            <thead>
+                                <tr>
+                                    <th style="text-align:center"> Name </th>
```
wluyima: And all these labels need to be internationalized

omod/src/main/java/org/openmrs/web/controller/user/UserFormController.java
bholagabbar: After I believe. Just before the page redirects

omod/src/main/webapp/admin/users/userForm.jsp
bholagabbar: Meaning? Add them to messages file in core?

omod/src/main/webapp/admin/users/userForm.jsp
```diff
@@ -103,9 +103,42 @@
...
+                                    <th style="text-align:center"> Identifier </th>
+                                </tr>
+                            </thead>
+                            <tbody>
+                                <c:forEach items="${providerList}" var="currentProvider">    
+                                    <tr>
+                                       <td align="center"> <c:out value="${currentProvider.getName()}"/> </td>
```
wluyima: currentProvider.name is the cleaner way to instead of calling the getter in the gsp

omod/src/main/webapp/admin/users/userForm.jsp
```diff
@@ -103,9 +103,42 @@
...
+                                       <td align="center"> <c:out value="${currentProvider.getIdentifier()}"/> </td>
```
wluyima: same here currentProvider.identifier should be enough

omod/src/main/webapp/admin/users/userForm.jsp
```diff
@@ -103,9 +103,42 @@
...
+                                    </tr>
+                                </c:forEach>
+                            </tbody>
+                        </table>
+                    </c:if>
+                    
+                    <c:if test="${isProvider == false}">
+                        <input name="providerCheckBox" type="checkbox" value="addToProviderTable"/>
+                        <label for="providerCheckBox">Create a Provider account for this User</label>
```
wluyima: Translate this

omod/src/main/webapp/admin/users/userForm.jsp
bholagabbar: Have a look at the ticket comments and images. It is explained there. Basically, there are 2 possibilities. If the user is new, we display a check box to add him to provider table 

Else, we display a table of all the providers the person associated with the user is.

omod/src/main/webapp/admin/users/userForm.jsp
bholagabbar: Ok

omod/src/main/java/org/openmrs/web/controller/user/UserFormController.java
bholagabbar: I understand your concern as all previous error checks should be met. It is after. No worries

omod/src/test/java/org/openmrs/web/controller/user/UserFormControllerTest.java
```diff
@@ -31,18 +34,38 @@
...
+	@Test
+	@Verifies(value = "Adds ", method = "handleSubmission(WebRequest,HttpSession,String,String,String,null, String, User,BindingResult)")
+	public void handleSubmission_addUserToProviderTableWhenProviderAndProviderTableOptionAreSelected() throws Exception {
+		WebRequest request = new ServletWebRequest(new MockHttpServletRequest());
+		User user = controller.formBackingObject(request, null);
+		user.addName(new PersonName("This", "is", "Test"));
+		user.getPerson().setGender("F");
+		user.getPerson().setId(1234);
+		ModelMap model = new ModelMap();
+		model.addAttribute("isProvider", false);
```
wluyima: You need an assertion around here that this person had no provider account before the form submission

omod/src/main/webapp/admin/users/userForm.jsp
bholagabbar: Alright will do that

omod/src/main/webapp/admin/users/userForm.jsp
bholagabbar: Have a look at my comment [here](https://github.com/openmrs/openmrs-module-legacyui/pull/41/files#discussion_r54492052) and on the ticket discussion between me and Darius.

omod/src/test/java/org/openmrs/web/controller/user/UserFormControllerTest.java
bholagabbar: Ok

omod/src/main/webapp/admin/users/userForm.jsp
bholagabbar: Hold on. Translate you mean in different languages?

omod/src/main/webapp/admin/users/userForm.jsp
bholagabbar: @wyclif I thought you were asking for me to explain the code lol. You mean translate this to different languages yes?

omod/src/main/java/org/openmrs/web/controller/user/UserFormController.java
bholagabbar: @wluyima I think it would make sense to add some JavaScript that automatically checks provider when this option is selected and unchecks it (if it was not checked previously) when unchecked

omod/src/main/webapp/admin/users/userForm.jsp
```diff
@@ -103,9 +103,42 @@
...
+                                    <tr>
+                                       <td align="center"> <c:out value="${currentProvider.getName()}"/> </td>
+                                       <td align="center"> <c:out value="${currentProvider.getIdentifier()}"/> </td>
+                                    </tr>
+                                </c:forEach>
+                            </tbody>
+                        </table>
+                    </c:if>
+                    
+                    <c:if test="${isProvider == false}">
```
wluyima: I think ${isProvider} should be fine

omod/src/main/java/org/openmrs/web/controller/user/UserFormController.java
wluyima: And what I'm trying to saying is that in future it's good practice to use a boolean value for a checkbox instead of a String in the HTML like you did in the JSP, that way the type conversion to boolean is automatic on the server side

omod/src/main/java/org/openmrs/web/controller/user/UserFormController.java
```diff
@@ -284,6 +290,13 @@ public String handleSubmission(WebRequest request, HttpSession httpSession, Mode
 				us.changeQuestionAnswer(user, secretQuestion, secretAnswer);
 			}
 			
+			//Check if admin wants the person associated with the user to be added to the Provider Table
+			if(addToProviderTableOption != null) {
+				Provider provider = new Provider();
+				provider.setPerson(user.getPerson());
+				provider.setIdentifier(user.getSystemId());
```
wluyima: Was this agreed on, that we should default to systemId as the identifier? We should have tried to emulate the ref app where we prompt the user for an identifier

bholagabbar: @wluyima Yeah. The system id is generated by the system automatically and acts as a good Identifier. The person who reported the issue from AMPATH specifically requested that we add the system id as identifier. You can find id in the description

omod/src/main/java/org/openmrs/web/controller/user/UserFormController.java
bholagabbar: @wluyima So do I let it be or change it? You're saying that if I change it to `Boolean`, it is automatically converted?

omod/src/main/java/org/openmrs/web/controller/user/UserFormController.java
```diff
@@ -125,6 +128,24 @@ public String showForm(@RequestParam(required = false, value = "userId") Integer
 			model.addAttribute("secretQuestion", userService.getSecretQuestion(user));
 		}
 		
+		if (user.getPerson().getId() != null && !Context.getProviderService().getProvidersByPerson(user.getPerson()).isEmpty()) {
+			model.addAttribute("isProvider", true);
+			
+			//Since identifier can be null, check corresponding identifiers and replace the 'nulls' with a "no identifier"
+			List<Provider> providerList = new ArrayList<Provider>(Context.getProviderService().getProvidersByPerson(user.getPerson()));
```
bholagabbar: @wluyima I had to use a `List` since you can't access the contents of a `Collection`.

api/src/main/resources/messages.properties
```diff
@@ -1,3 +1,5 @@
 ${project.parent.artifactId}.title=Legacy UI Module
 ${project.parent.artifactId}.manage=Manage module
 ${project.parent.artifactId}.patient.duplicateName.voided={0} is a duplicate name for the same patient and it's voided. Please restore the existing name instead of creating new name
+User.providerAccount=Provider Account
```
wluyima: prefix the codes with the module id just like the existing ones

omod/src/main/java/org/openmrs/web/controller/user/UserFormController.java
wluyima: A collection has an iterator you can use to traverse its elements, so you don't need to convert it to a list

wluyima: Do you mean that in the JSP you can't iterate over the collection or you do you mean you had no way to iterate over the collection here in the controller?

omod/src/main/java/org/openmrs/web/controller/user/UserFormController.java
```diff
@@ -125,6 +128,24 @@ public String showForm(@RequestParam(required = false, value = "userId") Integer
...
+			for (int i = 0; i < providerList.size(); i++) {
+				if (providerList.get(i).getIdentifier() == null) {
+					Provider currentProvider = providerList.get(i);
+					currentProvider.setIdentifier("no identifier");
```
wluyima: This is actually bad practice, you don't want to set a new property value especially when you're dealing with hibernate managed objects because these are normally proxies, also **No identifier** needs to be translated, **No Identifier Specified** would probably be clearer. The proper thing to do is not set it here and in the JSP you check if it's not null you print the value otherwise display the 'No Identifier Specified' text, it would also be nice to put it in bold and italics so that one doesn't thing it is the actual identifier but rather a place holder

omod/src/main/java/org/openmrs/web/controller/user/UserFormController.java
wluyima: Since it works the way it is, It's really up to you if you want to change, but it's better style to use Boolean for a single checkbox

omod/src/main/webapp/admin/users/userForm.jsp
```diff
@@ -103,9 +103,25 @@
 			</tr>
 		</table>
 	</fieldset>	
-	
 	<br/>
 	
+    <fieldset>
+        <legend><openmrs:message code="User.providerAccount"/></legend>
+        <c:if test="${isProvider}">
+            Provider Identifier(s):
```
wluyima: You forgot to translate this

omod/src/main/java/org/openmrs/web/controller/user/UserFormController.java
bholagabbar: Aah silly me. No a collection works fine I will change it. A list is just more convenient in an enhanced for loop.

api/src/main/resources/messages.properties
bholagabbar: I didn't understand how the prefix codes work. Can you please explain how I'm supposed to do it? And why?

omod/src/main/webapp/admin/users/userForm.jsp
bholagabbar: Yeah my bad. I'll do that

omod/src/main/java/org/openmrs/web/controller/user/UserFormController.java
bholagabbar: @wluyima Anyways, I just need to pass this over to the JSP. I'll check for the null values there and print the message "No Identifier Specified" as you asked me to

api/src/main/resources/messages.properties
```diff
@@ -1,3 +1,8 @@
 ${project.parent.artifactId}.title=Legacy UI Module
 ${project.parent.artifactId}.manage=Manage module
 ${project.parent.artifactId}.patient.duplicateName.voided={0} is a duplicate name for the same patient and it's voided. Please restore the existing name instead of creating new name
+
+legacyui.manageuser.providerAccount=Provider Account
```
bholagabbar: I spoke to Darius on IRC regarding prefixing module names and we mutually agreed to keep it this way. Hope it's alright now

omod/src/main/webapp/admin/users/userForm.jsp
```diff
@@ -103,9 +103,31 @@
...
 	</fieldset>	
-	
 	<br/>
 	
+    <fieldset>
+        <legend><openmrs:message code="legacyui.manageuser.providerAccount"/></legend>
+        <c:if test="${isProvider}">
+            <openmrs:message code="legacyui.manageuser.providerIdentfier"/>
+            <c:forEach items="${providerList}" var="currentProvider" varStatus="loop">   
+                <c:if test="${empty currentProvider.identifier}">
```
bholagabbar: `empty` checks for null values and prints the value accordingly. I believe this is how you wanted it done in the JSP :)

api/src/main/resources/messages.properties
wluyima: This is what i was suggesting, so it looks fine

omod/src/test/java/org/openmrs/web/controller/user/UserFormControllerTest.java
```diff
@@ -31,18 +34,39 @@
...
+		user.addName(new PersonName("This", "is", "Test"));
+		user.getPerson().setGender("F");
+		user.getPerson().setId(1234);
+		Assert.assertTrue(Context.getProviderService().getProvidersByPerson(user.getPerson()).isEmpty());
+		ModelMap model = new ModelMap();
+		model.addAttribute("isProvider", false);
+		controller.showForm(1234, "true", user, model);
+		controller.handleSubmission(request, new MockHttpSession(), new ModelMap(), "", "Save User", "pass123", "pass123", null,
+				null, null, new String[0] , "true", "addToProviderTable", user, new BindException(user, "user"));
+		Assert.assertNotNull(Context.getProviderService().getProvidersByPerson(user.getPerson()));
```
wluyima: This should be checking that the list is not empty, ProviderService.getProvidersByPerson can never return null

bholagabbar:
@wluyima Great! Can we merge and get this over with? We will also have to simultaneously backport this to 1.11.x and 1.12.x. Ping me and I will issue request for them ASAP so we can closed this as fixed for AMPATH :)

wluyima:
There is still a comment that's not responded to about an incorrect test and it should be ready to go

omod/src/test/java/org/openmrs/web/controller/user/UserFormControllerTest.java
```diff
@@ -22,27 +24,50 @@
...
+	}
+	
+	/**
+	 * @see UserFormController#handleSubmission(WebRequest,HttpSession,String,String,String,null, String, User,BindingResult)
+	 *
+	 */
+	@Test
+	@Verifies(value = "Creates Provider Account when Provider Account Checkbox is selected", method = "handleSubmission(WebRequest,HttpSession,String,String,String,null, String, User,BindingResult)")
+	public void handleSubmission_createUserProviderAccountWhenProviderAccountCheckboxIsSelected() throws Exception {
+		executeDataSet(TEST_DATA);
```
wluyima: FYI, you could've used one of the test users in the standardDataset.xml file without creating a new one

bholagabbar:
@wluyima I created the PR for core. here: https://github.com/openmrs/openmrs-core/pull/1731

omod/src/main/webapp/admin/users/userForm.jsp
wyclif: Hello, my name is Wyclif.

PR #43, opened 2016-04-29
Add tests for PatientFormController and sonar/jacoco to pom.xml

ldf92:
Add tests for PatientFormController.

Add sonar/jacoco plugins to pom.xml so that we can get coverage information.


wluyima:
What is the ticket number?

dkayiwa:
@ldf92 what ticket number is this?
I would also use a separate pull request for the sonar changes instead of mixing them with those of the unit tests.

teleivo:
just a little remark. you are using the old test style. the AT verifies annotation is now only in the javadoc of the test, see https://wiki.openmrs.org/display/docs/Generate+Test+Case+Plugin

dkayiwa:
@ldf92 create new pull requests that separate the two.

PR #48, opened 2016-06-30
LUI-78 Enable use of uuid as patientId param in patientDashboard

teleivo:
PatientDashboardController can now be reached with request param patientId values of patientId and UUID
Enables developers using REST where UUID as common to navigate to this page

* extract getPatient method for patient id and uuid
* add test class

see https://issues.openmrs.org/browse/LUI-78


teleivo:
via existing patient ID
image

via existing patient UUID
image

with non existing patient ID
image

with non existing patient UUID
image

with blank
image

PR #49, opened 2016-06-30
Add RegexValidatedTextDatatypeHandler

teleivo:
RegexValidatedTextDatatype is not validated for when setting a global property
datatype to it since now handler exists. This prevents using regex for
validation of global properties.

* added RegexValidatedTextDatatypeHandler with tests

see https://issues.openmrs.org/browse/LUI-81

### DIFF
--- a/omod/src/main/java/org/openmrs/web/attribute/handler/RegexValidatedTextDatatypeHandler.java
+++ b/omod/src/main/java/org/openmrs/web/attribute/handler/RegexValidatedTextDatatypeHandler.java
@@ -1,0 +1,90 @@
+/**
+ * This Source Code Form is subject to the terms of the Mozilla Public License,
+ * v. 2.0. If a copy of the MPL was not distributed with this file, You can
+ * obtain one at http://mozilla.org/MPL/2.0/. OpenMRS is also distributed under
+ * the terms of the Healthcare Disclaimer located at http://openmrs.org/license.
+ * Copyright (C) OpenMRS Inc. OpenMRS is a registered trademark and the OpenMRS
+ * graphic logo is a trademark of OpenMRS Inc.
+ */
+package org.openmrs.web.attribute.handler;
+
+import java.util.Map;
+
+import javax.servlet.http.HttpServletRequest;
+
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
+import org.openmrs.api.context.Context;
+import org.openmrs.customdatatype.CustomDatatype;
+import org.openmrs.customdatatype.InvalidCustomValueException;
+import org.openmrs.customdatatype.datatype.RegexValidatedTextDatatype;
+import org.springframework.stereotype.Component;
+
+/**
+ * A handler for the {@link RegexValidatedTextDatatype}.
+ */
+@Component
+public class RegexValidatedTextDatatypeHandler implements FieldGenDatatypeHandler<RegexValidatedTextDatatype, java.lang.String> {
+	
+	
+	private static final Log log = LogFactory.getLog(RegexValidatedTextDatatypeHandler.class);
+	
+	/**
+	 * @see org.openmrs.customdatatype.CustomDatatypeHandler#setHandlerConfiguration(String)
+	 */
+	@Override
+	public void setHandlerConfiguration(String arg0) {
+		// not used
+	}
+	
+	/**
+	 * @see org.openmrs.web.attribute.handler.FieldGenDatatypeHandler#getWidgetName()
+	 */
+	@Override
+	public String getWidgetName() {
+		return "java.lang.String";
+	}
+	
+	/**
+	 * @see org.openmrs.web.attribute.handler.FieldGenDatatypeHandler#getWidgetConfiguration()
+	 */
+	@Override
+	public Map<String, Object> getWidgetConfiguration() {
+		return null;
+	}
+	
+	/**
+	 * @see org.openmrs.web.attribute.handler.FieldGenDatatypeHandler#getValue(CustomDatatype,
+	 *      HttpServletRequest, String)
+	 */
+	@Override
+	public String getValue(RegexValidatedTextDatatype datatype, HttpServletRequest request, String formFieldName)
+	        throws InvalidCustomValueException {
+		String stringVal = request.getParameter(formFieldName);
+		try {
+			datatype.validate(stringVal);
+		}
+		catch (InvalidCustomValueException ex) {
+			throw new InvalidCustomValueException("Invalid value: " + stringVal);
+		}
+		return stringVal;
+	}
+	
+	/**
+	 * @see org.openmrs.web.attribute.handler.HtmlDisplayableDatatypeHandler#toHtmlSummary(CustomDatatype,
+	 *      String)
+	 */
+	@Override
+	public CustomDatatype.Summary toHtmlSummary(CustomDatatype<String> datatype, String valueReference) {
+		return new CustomDatatype.Summary(toHtml(datatype, valueReference), true);
+	}
+	
+	/**
+	 * @see org.openmrs.web.attribute.handler.HtmlDisplayableDatatypeHandler#toHtml(CustomDatatype,
+	 *      String)
+	 */
+	@Override
+	public String toHtml(CustomDatatype<String> datatype, String valueReference) {
+		return valueReference;
+	}
+}
--- a/omod/src/test/java/org/openmrs/web/attribute/handler/RegexValidatedTextDatatypeHandlerTest.java
+++ b/omod/src/test/java/org/openmrs/web/attribute/handler/RegexValidatedTextDatatypeHandlerTest.java
@@ -1,0 +1,118 @@
+/**
+ * This Source Code Form is subject to the terms of the Mozilla Public License,
+ * v. 2.0. If a copy of the MPL was not distributed with this file, You can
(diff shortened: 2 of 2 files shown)


teleivo:
as a test I added a GP with following settings:
image

and tried to set an invalid value on the GP

image

teleivo:
am open to suggestions for changes to `toHtml/toHtmlSummary`

teleivo:
by the way it also works without setting the `preferred_handler` of the global property, since the `RegexValidatedTextDatatypeHandler` is a Component and the only handler for the `RegexValidatedTextDatatype` it gets used by default.

omod/src/main/java/org/openmrs/web/attribute/handler/RegexValidatedTextDatatypeHandler.java
```diff
@@ -0,0 +1,90 @@
...
+	/**
+	 * @see org.openmrs.web.attribute.handler.FieldGenDatatypeHandler#getValue(CustomDatatype,
+	 *      HttpServletRequest, String)
+	 */
+	@Override
+	public String getValue(RegexValidatedTextDatatype datatype, HttpServletRequest request, String formFieldName)
+	        throws InvalidCustomValueException {
+		String stringVal = request.getParameter(formFieldName);
+		try {
+			datatype.validate(stringVal);
```
wluyima: Validation should happen in the validate method of the datatype and not the handler

omod/src/main/java/org/openmrs/web/attribute/handler/RegexValidatedTextDatatypeHandler.java
```diff
@@ -0,0 +1,90 @@
...
+import org.openmrs.customdatatype.CustomDatatype;
+import org.openmrs.customdatatype.InvalidCustomValueException;
+import org.openmrs.customdatatype.datatype.RegexValidatedTextDatatype;
+import org.springframework.stereotype.Component;
+
+/**
+ * A handler for the {@link RegexValidatedTextDatatype}.
+ */
+@Component
+public class RegexValidatedTextDatatypeHandler implements FieldGenDatatypeHandler<RegexValidatedTextDatatype, java.lang.String> {
```
wluyima: Since this extends FieldGenDatatypeHandler, it should be in the legacy UI module

omod/src/main/java/org/openmrs/web/attribute/handler/RegexValidatedTextDatatypeHandler.java
```diff
@@ -0,0 +1,90 @@
...
+	public void setHandlerConfiguration(String arg0) {
+		// not used
+	}
+	
+	/**
+	 * @see org.openmrs.web.attribute.handler.FieldGenDatatypeHandler#getWidgetName()
+	 */
+	@Override
+	public String getWidgetName() {
+		return "java.lang.String";
```
wluyima: String it too generic there can be other datatype handlers that work with strings, I would name this something like  regularExpression for clarity

omod/src/main/java/org/openmrs/web/attribute/handler/RegexValidatedTextDatatypeHandler.java
```diff
@@ -0,0 +1,90 @@
...
+		return stringVal;
+	}
+	
+	/**
+	 * @see org.openmrs.web.attribute.handler.HtmlDisplayableDatatypeHandler#toHtmlSummary(CustomDatatype,
+	 *      String)
+	 */
+	@Override
+	public CustomDatatype.Summary toHtmlSummary(CustomDatatype<String> datatype, String valueReference) {
+		return new CustomDatatype.Summary(toHtml(datatype, valueReference), true);
```
wluyima: This should typically call
`toHtmlSummary(datatype, valueReference).getSummary();`

omod/src/main/java/org/openmrs/web/attribute/handler/RegexValidatedTextDatatypeHandler.java
```diff
@@ -0,0 +1,90 @@
...
+	}
+	
+	/**
+	 * @see org.openmrs.web.attribute.handler.HtmlDisplayableDatatypeHandler#toHtml(CustomDatatype,
+	 *      String)
+	 */
+	@Override
+	public String toHtml(CustomDatatype<String> datatype, String valueReference) {
+		return valueReference;
```
wluyima: This should call
`datatype.getTextSummary(valueReference);`

omod/src/main/java/org/openmrs/web/attribute/handler/RegexValidatedTextDatatypeHandler.java
wluyima: Basically, just get rid of the validation logic here, the framework automatically will call the validate method on the datatype

omod/src/main/java/org/openmrs/web/attribute/handler/RegexValidatedTextDatatypeHandler.java
wluyima: Sorry, I just realized this is already in the module, ignore my comment above

omod/src/main/java/org/openmrs/web/attribute/handler/RegexValidatedTextDatatypeHandler.java
teleivo: @wluyima if I do as you say, I get

image

because there is no such type. Looking for example to the [ProviderFieldGenDatatypeHandler ](ProviderFieldGenDatatypeHandler) it is "org.openmrs.Provider" and there is a fieldgen tag for the provider.

omod/src/main/java/org/openmrs/web/attribute/handler/RegexValidatedTextDatatypeHandler.java
teleivo: it is done by [DateFieldGenDatatypeHandler ](https://github.com/teleivo/openmrs-module-legacyui/blob/03e0c53d56b24c43e8304e56e53469a07da2946a/omod/src/main/java/org/openmrs/web/attribute/handler/DateFieldGenDatatypeHandler.java#L67-L69) thats where I got this from and its not necessarily wrong since I am allowed to throw a InvalidCustomValueException.

So currently it behaves like this when entering an invalid value according to the regex:
image

it also replaces the invalid value with the latest saved GP which was valid.


If I change to

```
@Override
	public String getValue(RegexValidatedTextDatatype datatype, HttpServletRequest request, String formFieldName)
	        throws InvalidCustomValueException {
		return request.getParameter(formFieldName);
	}
```

it becomes:

image

and leaves the invalid value.

@dkayiwa and @wluyima just tell me what you prefer and I can make another PR.

omod/src/main/java/org/openmrs/web/attribute/handler/RegexValidatedTextDatatypeHandler.java
teleivo: @wluyima but `RegexValidatedTextDatatypeHandler ` is implementing the interface `HtmlDisplayableDatatypeHandler ` so toHtmlSummary(datatype, valueReference) is only implemented in this class. look at the other handlers.

omod/src/main/java/org/openmrs/web/attribute/handler/RegexValidatedTextDatatypeHandler.java
teleivo: what I dont like about this is that now I dont see which GP is making the problem. I think what we have now is nicer for the user.

omod/src/main/java/org/openmrs/web/attribute/handler/RegexValidatedTextDatatypeHandler.java
teleivo: I can change this if needed, tried it and it works :)

omod/src/main/java/org/openmrs/web/attribute/handler/RegexValidatedTextDatatypeHandler.java
wluyima: I thought the widget name is an arbitrary value, can you confirm @djazayeri ?

djazayeri: I don't remember how this is supposed to work, but there must be an example
of it somewhere...

-Darius (by phone)

wluyima: I guess java.lang.String is fine

omod/src/main/java/org/openmrs/web/attribute/handler/RegexValidatedTextDatatypeHandler.java
wluyima: I guess then it's fine to leave it this way

omod/src/main/java/org/openmrs/web/attribute/handler/RegexValidatedTextDatatypeHandler.java
wluyima: This works too, but if you look at datatype.getTextSummary(), it pretty much does the same thing as you're doing here so I find it as code duplication, if you wanted to display a different value as html, then it would make sense to do what you're doing.

PR #56, opened 2016-09-01
LUI-95 errorhandler.jsp cannot find openmrs taglib

teleivo:
On error in a jsp file the errorhandler page should nicely handle exceptions
for better user experience, however the taglib include was pointing to the
older path when the legacy ui was still part of the core.

* changed taglib uri to same value as in the (omod/src/main/webapp/template/include.jsp)

see https://issues.openmrs.org/browse/LUI-95


teleivo:
after the change the error page shows like:

image

PR #63, opened 2016-09-09
LUI-93 Add link to patientDashboard in patientHeader portlet

teleivo:
make an anchor out of the patient name if not on the patientDashboardForm
so users can easily navigate to the patients dashboard.

see https://issues.openmrs.org/browse/LUI-93

### DIFF
--- a/omod/src/main/webapp/portlets/patientHeader.jsp
+++ b/omod/src/main/webapp/portlets/patientHeader.jsp
@@ -19,7 +19,15 @@
 
 <%-- Header showing preferred name, id, and treatment status --%>
 <div id="patientHeader" class="boxHeader${model.patientVariation}">
-<div id="patientHeaderPatientName"><c:out value="${model.patient.personName}" /></div>
+<div id="patientHeaderPatientName">
+<c:choose>
+<c:when test="${model.showPatientDashboardLink == true}" >
+  <a href="${pageContext.request.contextPath}/patientDashboard.form?patientId=${model.patient.id}"><c:out value="${model.patient.personName}" /></a></div>
+</c:when>
+<c:otherwise>
+  <c:out value="${model.patient.personName}" /></div>
+</c:otherwise>
+</c:choose>
 <div id="patientHeaderPreferredIdentifier">
 	<c:if test="${fn:length(model.patient.activeIdentifiers) > 0}">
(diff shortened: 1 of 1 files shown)


omod/src/main/webapp/portlets/patientHeader.jsp
```diff
@@ -19,7 +19,15 @@
 
 <%-- Header showing preferred name, id, and treatment status --%>
 <div id="patientHeader" class="boxHeader${model.patientVariation}">
-<div id="patientHeaderPatientName"><c:out value="${model.patient.personName}" /></div>
+<div id="patientHeaderPatientName">
+<c:choose>
+<c:when test="${not fn:contains(pageContext.request.servletPath,'patientDashboardForm.jsp')}" >
```
wluyima: Since this is a portlet, I would think it's better to add a parameter/attribute to it that determines if the link should be displayed or not which defaults to false

teleivo:
@wluyima now it will only be added if included like so

```
<openmrs:portlet url="patientHeader" patientId="${patient.patientId}" parameters="showPatientDashboardLink=true" />
```

and not on the patientDashboardForm

omod/src/main/webapp/portlets/patientHeader.jsp
```diff
@@ -19,7 +19,15 @@
 
 <%-- Header showing preferred name, id, and treatment status --%>
 <div id="patientHeader" class="boxHeader${model.patientVariation}">
-<div id="patientHeaderPatientName"><c:out value="${model.patient.personName}" /></div>
+<div id="patientHeaderPatientName">
+<c:choose>
+<c:when test="${model.showPatientDashboardLink == true && not fn:endsWith(pageContext.request.servletPath,'patientDashboardForm.jsp')}" >
```
wluyima: Now that you're using showPatientDashboardLink parameter then you shouldn't have to check of the URL ends with patientDashboardForm.jsp anymore

teleivo:
@wluyima updated :+1:

teleivo:
@wluyima build fails due to access not being permitted to nexus repo see https://travis-ci.org/openmrs/openmrs-module-legacyui/builds/159346976#L195

not my change since it only affects the jsp.

PR #71, opened 2016-12-02
LUI-108 | Ability to search for patients to merge by identifiers

pankajladhar:
Ability to search for patients to merge by identifiers

### DIFF
--- a/omod/src/main/java/org/openmrs/web/dwr/DWRPatientService.java
+++ b/omod/src/main/java/org/openmrs/web/dwr/DWRPatientService.java
@@ -55,7 +55,7 @@ public class DWRPatientService implements GlobalPropertyListener {
 	private static final Log log = LogFactory.getLog(DWRPatientService.class);
 	
 	private static Integer maximumResults;
-	
+
 	/**
 	 * Search on the <code>searchValue</code>. If a number is in the search string, do an identifier
 	 * search. Else, do a name search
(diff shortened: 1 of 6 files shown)


omod/src/main/java/org/openmrs/web/dwr/DWRPatientService.java
```diff
@@ -335,7 +335,29 @@ public PatientListItem getPatient(Integer patientId) {
...
+	public Vector<Object> findPatientsByIdentifier(String[] identifiers) {
+		Vector<Object> patientList = new Vector<>();
+		for (String identifier : identifiers) {
+			try {
+				List<Patient> patients = Context.getPatientService().getPatients(identifier, false, 0, 1);
+				if(patients.size() > 0){
+					patientList.add(new PatientListItem(patients.get(0)));
+				}
+			} catch (Exception e) {
+				log.error("Error while searching for patients", e);
```
dkayiwa: Don't you wanna continue the loop and add more patients, just in case only a few fail?

omod/src/main/java/org/openmrs/web/dwr/DWRPatientService.java
```diff
@@ -335,7 +335,29 @@ public PatientListItem getPatient(Integer patientId) {
...
+				return new Vector<>();
```
dkayiwa: Wouldn't you just return patientList instead of doing another instantiation?

omod/src/main/java/org/openmrs/web/dwr/DWRPatientService.java
pankajladhar: We have implemented the given feedback.
To reduce search time, we have used `getPatientIdentifiers()` instead of `getPatients` to find a patient.

PR #79, opened 2017-03-25
Refactor all request:parameter tags

isears:


isears:
I couldn't find any existing JS-escaping functions so I created my own and added them to the taglib. Otherwise, most of these changes are pretty straightforward refactors of request:parameter

omod/src/main/webapp/admin/concepts/conceptDrugForm.jsp
```diff
@@ -10,9 +10,8 @@
 <script type="text/javascript">
 	dojo.require("dojo.widget.openmrs.ConceptSearch");
 	dojo.require("dojo.widget.openmrs.OpenmrsPopup");
-
 	<request:existsParameter name="autoJump">
-		var autoJump = <request:parameter name="autoJump"/>;
+		var autoJump = ${openmrs:getSafeJsBoolean(param.autoJump)};
```
dkayiwa: Wouldn't <c:out value work here instead of the new tag?

isears: I tried that, but I found that <c:out> does some html-encoding that will cause the JS to break for certain inputs. Since the value is injected straight into JS code (and not HTML) we really need a JS-specific filtering function.

dkayiwa: Do you mind giving an example? 😊

isears: Yeah, no problem

Take this request:
`conceptDrug.form?autoJump=test'alert(1)`

If the JSP were using <c:out> tags, `test'alert(1)` would turn into `test&#39;alert(1)` and would generate the following JS:
`var autoJump = test&#39;alert(1);`

dkayiwa: Would that execute the alert? Or result into a javascript error?

isears: I can't remember if I ever actually created the alert popup. I know that the specific example I outlined above will at least break the page's javascript on chrome. I'm almost certain I could develop  a payload that would circumvent c:out filtering to execute the alert, but it would take me another day or so.

dkayiwa: Shouldn't a javascript error be fine for someone trying to do a JavaScript attack?

isears: Yeah, I'd say even just the error is a security concern.

dkayiwa: How? 😊

isears: I'm sorry, I think I misunderstood your first comment.

Either way, <c:out> really isn't meant for escaping strings that will be used as javascript because javascript doesn't recognize html encoding. I don't have access to my development environment right now, but in a few hours I'll try to put together a proof-of-concept for <c:out> circumvention and get back to you on this.

isears: @dkayiwa the following url will trigger an alert message when using the <c:out> tag. I tested on the latest version of firefox:
`http://localhost:8080/openmrs/admin/concepts/conceptDrug.form?autoJump=false;alert(1);`

However, during testing I realized there was a bug in my filtering function (oops!) so I'm going to close this PR and re-open with revised changes.

PR #80, opened 2017-03-27
Refactor all request:parameter tags

isears:


omod/src/main/java/org/openmrs/web/taglib/functions/Util.java
```diff
@@ -0,0 +1,67 @@
...
+	
+	/**
+	 * This method will make untrusted strings safe for use as JavaScript integers
+	 *
+	 * @param s
+	 * @return either the original value (if an int) or 'null'
+	 */
+	public static String getSafeJsInt(String s){
+		for(int i = 0; i < s.length(); i++){
+			if(!Character.isDigit(s.charAt(i))){
```
dkayiwa: How about using something like org.apache.commons.lang.math.NumberUtils.isNumber()

omod/src/main/java/org/openmrs/web/taglib/functions/Util.java
```diff
@@ -0,0 +1,67 @@
...
+	}
+	
+	/**
+	 * This method will make untrusted strings safe for use as JavaScript booleans
+	 *
+	 * @param s
+	 * @return one of 'true', 'false', or 'null'
+	 */
+	public static String getSafeJsBoolean(String s){
+		if(s.equals("true") || s.equals("false") || s.equals("0") || s.equals("1")){
```
dkayiwa: How about BooleanUtils.toBooleanObject()?

isears: Yeah that would be better. I will submit changes shortly (for both functions).

dkayiwa:
Thanks @isears 👍

PR #83, opened 2017-11-27
LUI-133: Fix error on module load when module uses Spring-injected filters

kwburnett:
Ensuring the Spring context has been refreshed before filters get loaded in case filters need dependencies defined in the Spring context. Previously, filters would be loaded before the Spring context was refreshed. This would lead to issues when the filters had a class of org.springframework.web.filter.DelegatingFilterProxy, meaning Spring is supposed to resolve all dependencies of the filter. If the context had not been refreshed, no beans would be found.

Additionally, an issue was also occurring when no "dwr-invoker" servlet was present, so made the reInitServlet only called if a servlet was present.

See this discussion on OpenMRS Talk for more clarification.
https://talk.openmrs.org/t/loading-modules-error-due-to-filters-and-spring-beans/14162

Here is the link to the JIRA item.
https://issues.openmrs.org/browse/LUI-133

PR #86, opened 2017-12-05
LUI-133: Fix error on module load when module uses Spring-injected filters

kwburnett:
Ensuring the Spring context has been refreshed before filters get loaded in case filters need dependencies defined in the Spring context. Previously, filters would be loaded before the Spring context was refreshed. This would lead to issues when the filters had a class of org.springframework.web.filter.DelegatingFilterProxy, meaning Spring is supposed to resolve all dependencies of the filter. If the context had not been refreshed, no beans would be found.

Additionally, an issue was also occurring when no "dwr-invoker" servlet was present, so made the reInitServlet only called if a servlet was present.

See this discussion on OpenMRS Talk for more clarification.
https://talk.openmrs.org/t/loading-modules-error-due-to-filters-and-spring-beans/14162

Here is the link to the JIRA item.
https://issues.openmrs.org/browse/LUI-133

This is a continuation of https://github.com/openmrs/openmrs-module-legacyui/pull/83.


dkayiwa:
@kwburnett can you also update the JIRA ticket with the new pull request link.

kwburnett:
@dkayiwa done.

PR #89, opened 2018-01-09
[LUI-66] Concept Proposal Form is broken

Sub6Resources:
See https://issues.openmrs.org/browse/LUI-66

### DIFF
--- a/omod/src/main/webapp/admin/concepts/conceptProposalForm.jsp
+++ b/omod/src/main/webapp/admin/concepts/conceptProposalForm.jsp
@@ -133,15 +133,7 @@
 									title="<c:out value="${listItem.description}"/>"
 									class='searchHit'>
 									${status.index + 1})
-									<c:choose >
-										<c:when test="${listItem.synonym != ''}">
-											<span class='mainHit'>${listItem.synonym}</span>
-											<span class='additionalHit'>&rArr; <c:out value="${listItem.name}"/></span>
-										</c:when>
-										<c:otherwise>
-											<span class='mainHit'><c:out value="${listItem.name}"/></span>
-										</c:otherwise>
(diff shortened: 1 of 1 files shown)


omod/src/main/webapp/admin/concepts/conceptProposalForm.jsp
```diff
@@ -134,8 +134,8 @@
 									class='searchHit'>
 									${status.index + 1})
 									<c:choose >
-										<c:when test="${listItem.synonym != ''}">
-											<span class='mainHit'>${listItem.synonym}</span>
+										<c:when test="${listItem.description != ''}">
```
dkayiwa: From here, https://github.com/openmrs/openmrs-core/blob/1.12.x/web/src/main/java/org/openmrs/web/dwr/ConceptListItem.java#L46-L53 the synonym was deprecated and hence removed. Therefore, just remove the choose and simply use listItem.name which is already in the otherwise section.

Sub6Resources: Thank you. I was viewing the wrong ConceptListItem.java so didn't find this deprecated statement.

dkayiwa: No. You did not look at the wrong guy. You just looked after it was deprecated and eventually removed. 😊

Sub6Resources:
The changes have been made @dkayiwa

omod/src/main/webapp/admin/concepts/conceptProposalForm.jsp
Sub6Resources: Ah, on an older version, that makes sense.

PR #101, opened 2019-01-28
Admin to add and delete account

ODORA0:
Admin is granted access to add and delete an account.
Issue worked on: https://issues.openmrs.org/browse/LUI-135


api/src/main/java/org/openmrs/module/legacyui/LegacyUIActivator.java
```diff
@@ -21,6 +21,8 @@
 public class LegacyUIActivator extends BaseModuleActivator {
 	
 	protected Log log = LogFactory.getLog(getClass());
+	
+
```
samuelmale: These are just whitespace changes, Are you still working on this

PR #12, opened 2015-10-29
LUI-14

jdegraft:
Changes following review
- correct file formatting

PR #103, opened 2019-02-12
LUI-136 : trimm off unnecesasry characters from the converted input s…

mozzy11:
Ticket : https://issues.openmrs.org/browse/LUI-136

in this commit , i trimmed off unnecesary characters , so that the converted input value can remain a pure Numeric string that can be converted to a conceptId


omod/src/main/java/org/openmrs/web/dwr/DWRConceptService.java
```diff
@@ -236,7 +236,11 @@
 	 * @param conceptId the id to look for
 	 * @return a {@link ConceptListItem} or null if conceptId is not found
 	 */
-	public ConceptListItem getConcept(Integer conceptId) {
```
dkayiwa: What is the reason behind changing from the integer conceptId to string phrase as the method parameter?

mozzy11: the issue is that,  the cause of the error was arising from the fact that the value to be taken as a conceptId here ` getConcept(Integer conceptId)` was in the form say "Concept #33" for id 33, "Concept #6" for id 6,, hence  say "Concept #66" couldnt be converted to a conceptid ,then hats why it was throwing the javascript error described in the ticket. 
So i changed from integer to string , so that the value  eg "Concept #66" be recieved as a string , then i trimm off the "Concept #" to remain with a Numeric string "66" , and then i convert it back to a concept id

mozzy11:
hello @dkayiwa ,  made a change in the PR, i think this is a better way

omod/src/main/webapp/tags/conceptField.tag
```diff
@@ -76,6 +77,10 @@
 		
 	})
 	
+	function convert(conc){
```
dkayiwa: What is this function doing?

mozzy11: it converts the  string(concept object) eg "concept #12"  to a numeric string (concept id) eg "12"

PR #108, opened 2019-03-08
LUI-123: Capability of searching Concepts by Concept UUIDs

Reyano132:
https://issues.openmrs.org/browse/LUI-123

Added the capability of searching Concepts by Concept UUIDs


Reyano132:
Yes, we can search the concept by its UUID. We just need to change the value of Concept.search from messages.properties file for changing the label of search box

Reyano132:
Hey @Ruhanga , I added unit tests for search Concepts by uuid. All the test cases are similar to test case made for search Concepts by id as id and uuid both are unique constraints of Concepts.

ibacher (approved):

PR #109, opened 2019-04-14
LUI-80: .../openmrs/options.form should have one field for Old Passwo…

ODORA0:
…rd and dispay right validation message for it when submitted as
Ticket ID: https://issues.openmrs.org/browse/LUI-80
Issue worked on: Removed one old password field, and changed output to enter password rather than password incorrect as explained in the ticket description

### DIFF
--- a/omod/src/main/java/org/openmrs/web/controller/OptionsFormController.java
+++ b/omod/src/main/java/org/openmrs/web/controller/OptionsFormController.java
@@ -182,13 +182,7 @@ public class OptionsFormController extends SimpleFormController {
 				catch (APIException e) {
 					errors.rejectValue("oldPassword", "error.password.match");
 				}
-			} else {
-				// if they left the old password blank but filled in new
-				// password
-				if (!"".equals(opts.getNewPassword())) {
-					errors.rejectValue("oldPassword", "error.password.incorrect");
-				}
-			}
+			} 
 			
 			if (!"".equals(opts.getSecretQuestionPassword())) {
 				if (!errors.hasErrors()) {
@@ -203,7 +197,7 @@ public class OptionsFormController extends SimpleFormController {
 			} else if (!"".equals(opts.getSecretAnswerNew())) {
 				// if they left the old password blank but filled in new
 				// password
-				errors.rejectValue("secretQuestionPassword", "error.password.incorrect");
+				errors.rejectValue("secretQuestionPassword", "Enter old password");
 			}
 			
 			String notifyType = opts.getNotification();


dkayiwa:
Can you attach screenshots on the JIRA ticket with the effects of your changes as advised at https://wiki.openmrs.org/display/docs/Pull+Request+Tips

omod/src/main/java/org/openmrs/web/controller/OptionsFormController.java
```diff
@@ -182,13 +182,7 @@ protected ModelAndView onSubmit(HttpServletRequest request, HttpServletResponse
 				catch (APIException e) {
 					errors.rejectValue("oldPassword", "error.password.match");
 				}
-			} else {
-				// if they left the old password blank but filled in new
-				// password
-				if (!"".equals(opts.getNewPassword())) {
```
dkayiwa: What is the reason behind removing these lines?

omod/src/main/java/org/openmrs/web/controller/OptionsFormController.java
```diff
@@ -203,7 +197,7 @@ protected ModelAndView onSubmit(HttpServletRequest request, HttpServletResponse
 			} else if (!"".equals(opts.getSecretAnswerNew())) {
 				// if they left the old password blank but filled in new
 				// password
-				errors.rejectValue("secretQuestionPassword", "error.password.incorrect");
```
dkayiwa: What is the reason behind changing this from error.password.incorrect to error.password.enter?

omod/src/main/java/org/openmrs/web/controller/OptionsFormController.java
ODORA0: I expected those lines to be the second password field not needed

dkayiwa: Did you confirm?

omod/src/main/java/org/openmrs/web/controller/OptionsFormController.java
ODORA0: I intended the output for that to be enter password rather than password incorrect

omod/src/main/java/org/openmrs/web/controller/OptionsFormController.java
ODORA0: No I did not.

mozzy11:
@ODORA0 , your PR seems to only to address only one requirement of the ticket . Are you sure this Pr also adtreses the UI change requirement of the ticket??. Can you attach a screen shot to confirm the change is implemented?

omod/src/main/java/org/openmrs/web/controller/OptionsFormController.java
ODORA0: @dkayiwa  I see that the code change here is not so effective after testing it via the RefApp, am going to change this

omod/src/main/java/org/openmrs/web/controller/OptionsFormController.java
ODORA0: As for this I also don't see it's effectiveness to remove one password field I'll also have to review my code changes made

ODORA0:
I have updated the ticket on the changes I have done though for the code on removing one of the password fields is what am not as certain of.

dkayiwa:
Which password field do you want to remove? And why?

ODORA0:
It says in the issue description that there are two password fields and removing just one of them would be preferable rather than having two fields where the user should enter their Old Password on this same page, since the user is entering the same value

dkayiwa:
@ODORA0 if you find this ticket complicated, you can take up this: https://issues.openmrs.org/browse/TRUNK-5413

ODORA0:
Yes @dkayiwa  thanks for that link too

PR #114, opened 2020-03-26
LUI-151:Logout Button not working in demo

gitcliff:
link:https://issues.openmrs.org/browse/LUI-151

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,9 +1,9 @@
 language: java
 jdk:
-   - oraclejdk8
+   - openjdk8
 script: mvn clean install --batch-mode
 matrix:
-  - jdk: oraclejdk8
+  - jdk: openjdk8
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -1,9 +1,9 @@
...
 jdk:
-   - oraclejdk8
+   - openjdk8
 script: mvn clean install --batch-mode
 matrix:
-  - jdk: oraclejdk8
+  - jdk: openjdk8
 branches:
   only:
     - master
```
ibacher: I completely agree with this change, but we should make it separately from this PR.

ibacher:
@gitcliff Thanks so much for this! It works perfectly for me. If we can just remove the extraneous change to the Travis config from this PR, I'll merge it!

PR #115, opened 2020-03-31
LUI-153: Footer to show optional extra data from runtime properties.

HerbertYiga:
https://issues.openmrs.org/browse/LUI-153

Footer shows an optional extra data from run time properties.
Also added a screen short on the ticket ID

### DIFF
--- a/api/src/main/resources/messages.properties
+++ b/api/src/main/resources/messages.properties
@@ -6,6 +6,7 @@ legacyui.manageuser.providerAccount=Provider Account
 legacyui.manageuser.createProviderAccount=Create a Provider account for this user
 legacyui.manageuser.providerIdentfier=Provider Identifier(s):
 legacyui.manageuser.noProviderIdentifier=No Identifier Specified
+legacyui.footer.extradata=Docker Image
 
 ${project.parent.artifactId}.Location.purgeLocation=Permanently Delete Location
 ${project.parent.artifactId}.Location.confirmDelete=Are you sure you want to delete this Location? It will be permanently removed from the system.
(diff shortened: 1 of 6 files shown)


omod/src/main/webapp/template/footerFull.jsp
```diff
@@ -45,6 +45,8 @@
 			
 			<span id="codeVersion"><openmrs:message code="footer.version"/>: ${openmrsVersion}</span>
 			
+			<span id="dockerImage"><openmrs:message code="legacyui.footer.extradata"/>: ${openmrsVersion}</span>
```
mks-d, line 48: This `code` isn't necessary.

api/src/main/resources/messages.properties
```diff
@@ -6,6 +6,7 @@ legacyui.manageuser.providerAccount=Provider Account
 legacyui.manageuser.createProviderAccount=Create a Provider account for this user
 legacyui.manageuser.providerIdentfier=Provider Identifier(s):
 legacyui.manageuser.noProviderIdentifier=No Identifier Specified
+legacyui.footer.extradata=Docker Image
```
mks-d, line 9: This message should go away.

omod/src/main/webapp/template/footerFull.jsp
mks-d, line 48: In the end you'll have something like this:
```html
<span id="extraData">${footerExtraData}</span>
```
However this span should be remain hidden if `${footerExtraData}` is blank or empty.

HerbertYiga:
@mks-d  okie,i have got the concept and let me make the changes

PR #122, opened 2020-07-06
LUI-161, Unable to edit the created alert in Users section

Am-Coder:
**Description of what I changed**

The already scheduled users need not be scheduled again.  So, had to remove the already scheduled users.

**Issue I worked On**
see https://issues.openmrs.org/browse/LUI-161

### DIFF
--- a/omod/src/main/java/org/openmrs/notification/web/controller/AlertFormController.java
+++ b/omod/src/main/java/org/openmrs/notification/web/controller/AlertFormController.java
@@ -119,6 +119,8 @@ public class AlertFormController extends SimpleFormController {
 						if (!userIds.contains(userId)) {
 							recipientsToRemove.add(recipient);
 						}
+						//To remove the users that have already been scheduled
+						userIds.remove(userId);
 					}
 				}
 				for (AlertRecipient ar : recipientsToRemove) {
--- a/omod/src/main/java/org/openmrs/notification/web/controller/AlertFormController.java
+++ b/omod/src/main/java/org/openmrs/notification/web/controller/AlertFormController.java
@@ -118,9 +118,10 @@ public class AlertFormController extends SimpleFormController {
 						Integer userId = recipient.getRecipient().getUserId();
 						if (!userIds.contains(userId)) {
 							recipientsToRemove.add(recipient);
+						} else {
+							//To remove the users that have already been scheduled
+							userIds.remove(userId);
 						}
-						//To remove the users that have already been scheduled
-						userIds.remove(userId);
 					}
 				}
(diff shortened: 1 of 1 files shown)


omod/src/main/java/org/openmrs/notification/web/controller/AlertFormController.java
```diff
@@ -119,6 +119,8 @@ protected ModelAndView processFormSubmission(HttpServletRequest request, HttpSer
 						if (!userIds.contains(userId)) {
 							recipientsToRemove.add(recipient);
 						}
+						//To remove the users that have already been scheduled
+						userIds.remove(userId);
```
HerbertYiga, line 123: @Am-Coder done on this,have you tested this with a running instance of openmrs?

Am-Coder: Yup, I have verified the working using OpenMRS SDK on both distribution and platform server. Also tested this on OpenMRS-core separately by making use of the generated omod of legacy-ui module.

dkayiwa: I find your addition confusing because it is under a block for removing all recipients not in the userIds list. Which has nothing to do with clearing the userIds list.

Am-Coder: **userIds** hold the id's of users that will be called for scheduling as shown [here](https://github.com/openmrs/openmrs-module-legacyui/blob/df54970ee7bde1c3c0913bc600a27398f8aef614/omod/src/main/java/org/openmrs/notification/web/controller/AlertFormController.java#L129) . In [this](https://github.com/openmrs/openmrs-module-legacyui/blob/df54970ee7bde1c3c0913bc600a27398f8aef614/omod/src/main/java/org/openmrs/notification/web/controller/AlertFormController.java#L117) loop where I have added the change, all the recipients of alert are the recipients that have already been scheduled. These id's need to be removed from **userIds** , otherwise you will try to persist the duplicate and hence raise an exception. 

That is why I added the change in the loop.

dkayiwa: Shouldn't it be in some sort of else clause?

Am-Coder: Else clause will be more appropriate. Made the changes.

Am-Coder: @dkayiwa , are some more changes needed for this PR ?

PR #125, opened 2020-09-02
TRUNK-4954:ObsService.saveObs should throw a validation exception if the obs and its encounter point to different persons

HerbertYiga:
Added a validation exception on saving an obs

Ticket ID:https://issues.openmrs.org/browse/TRUNK-4954

### DIFF
--- a/omod/src/main/java/org/openmrs/web/controller/observation/ObsFormController.java
+++ b/omod/src/main/java/org/openmrs/web/controller/observation/ObsFormController.java
@@ -13,6 +13,7 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.util.HashMap;
 import java.util.Map;
+import java.util.Objects;
 
 import javax.servlet.ServletException;
 import javax.servlet.http.HttpServletRequest;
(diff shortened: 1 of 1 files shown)


omod/src/main/java/org/openmrs/web/controller/observation/ObsFormController.java
```diff
@@ -140,7 +141,12 @@ protected ModelAndView onSubmit(HttpServletRequest request, HttpServletResponse
 						errors.reject("editReason", "Obs.edit.reason.empty");
 						return showForm(request, response, errors);
 					}
-					
+					if (obs.getPerson() != null && obs.getEncounter() != null && obs.getEncounter().getPatient() != null) {
```
HerbertYiga, line 144: @ibacher   hi ian,what do think of having this validation exception in legacy ui instead of core,i added a screen shot on the ticket Id,thanks cc @dkayiwa

ibacher: @HerbertYiga I think this check should be added to the [`ObsValidator`](https://github.com/openmrs/openmrs-core/blob/master/api/src/main/java/org/openmrs/validator/ObsValidator.java) rather than here. That way, it's also enforced inside the API.

HerbertYiga: @ibacher   Thanks ian,let me do the fix there

sherrif10:
@HerbertYiga  Are you still working on this ??, did you see travis failure

PR #135, opened 2021-03-02
LUI-177: Fix changed method parameters causing the patient program fe…

mhawila:
…ature to break on patient dashboard

### DIFF
--- a/omod/src/main/java/org/openmrs/web/dwr/DWRProgramWorkflowService.java
+++ b/omod/src/main/java/org/openmrs/web/dwr/DWRProgramWorkflowService.java
@@ -269,10 +269,10 @@ public class DWRProgramWorkflowService {
 		s.savePatientProgram(pp);
 	}
 	
-	public void voidLastState(Integer patientProgramId, String programWorkflowUuid, String voidReason) {
+	public void voidLastState(Integer patientProgramId, Integer programWorkflowId, String voidReason) {
 		ProgramWorkflowService s = Context.getProgramWorkflowService();
 		PatientProgram pp = s.getPatientProgram(patientProgramId);
-		ProgramWorkflow wf = s.getWorkflowByUuid(programWorkflowUuid);
+		ProgramWorkflow wf = pp.getProgram().getWorkflow(programWorkflowId);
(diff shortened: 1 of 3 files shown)


omod/src/main/java/org/openmrs/web/dwr/DWRProgramWorkflowService.java
```diff
@@ -269,10 +269,10 @@ public void changeToState(Integer patientProgramId, Integer programWorkflowId, I
 		s.savePatientProgram(pp);
 	}
 	
-	public void voidLastState(Integer patientProgramId, String programWorkflowUuid, String voidReason) {
+	public void voidLastState(Integer patientProgramId, Integer programWorkflowId, String voidReason) {
```
dkayiwa, line 272: Shouldn't we just overload this method? Just incase some body else is using it.

mhawila: I thought of doing so but I searched throughout the module I didn't find anywhere else other than the place where the feature is broken. Also I don't know how DWR works, wouldn't the two method look similar? Because even now when the workflow Id is passed, it is simply converted to String and passed along as UUID.

dkayiwa: Can you take a look at this? https://github.com/search?l=&q=org%3Aopenmrs+voidLastState&ref=advsearch&type=Code

mhawila: I have taken a look. If you closely you will see the difference between the signature of the method and the parameters which are passed where it is being called. By the way this is how the method was defined prior to it being changed due to method deprecation (see (https://github.com/openmrs/openmrs-core/blob/1.11.x/web/src/main/java/org/openmrs/web/dwr/DWRProgramWorkflowService.java#L231-L237). I think it was an oversight on the person who changed it to remove the depricated method (https://github.com/openmrs/openmrs-core/blob/40e13ad07a9f66e483c0ac6e19924376e1028bc6/api/src/main/java/org/openmrs/api/ProgramWorkflowService.java#L536-L547)

PR #136, opened 2021-03-02
Sanitize HTML in legacyui error div to avoid XSS- EMPT 54

jmmacdo4:
By adding cout to the error spans as well as adding htmlEscaping to the error div, I was able to remove the possibility for a cross site scripting attack on the implementation ID, Name, Passphrase, and Description fields.

@isears

### DIFF
--- a/omod/src/main/webapp/admin/concepts/conceptProposalList.jsp
+++ b/omod/src/main/webapp/admin/concepts/conceptProposalList.jsp
@@ -87,7 +87,7 @@
     						onclick="selectProposal('${conceptProposal.conceptProposalId}')"
     						onmouseover="mouseOver(this)" onmouseout="mouseOut(this)"> 
 							<td valign="top">${conceptProposal.encounter.encounterId}</td>
-							<td valign="top">${conceptProposal.originalText}</td>
+							<td valign="top"><c:out value='${conceptProposal.originalText}'/></td>
 							<td valign="top"><openmrs:format user="${conceptProposal.creator}"/></td>
 							<td valign="top">${conceptProposal.dateCreated}</td>
 							<td valign="top">${map.value}</td>


omod/src/main/webapp/admin/maintenance/implementationIdForm.jsp
```diff
@@ -30,7 +30,7 @@
 				<td style="white-space: nowrap">
 					<input type="text" value="${status.value}" name="${status.expression}" maxlength="20" size="8"/>
 					<c:if test="${status.errorMessage != ''}">
-						<span class="error">${status.errorMessage}</span>
+						<span class="error"><c:out value="${status.errorMessage}"/></span>
 					</c:if>
```
HerbertYiga, line 34: Do we expect cross site scripting in such cases <input type="text" name="${status.expression}" value="${status.value}" size="40" />

HerbertYiga:
do we expect cross site scripting in such cases <input type="text" name="${status.expression}" value="${status.value}" size="40" />

omod/src/main/webapp/admin/maintenance/implementationIdForm.jsp
```diff
@@ -30,7 +30,7 @@
 				<td style="white-space: nowrap">
 					<input type="text" value="${status.value}" name="${status.expression}" maxlength="20" size="8"/>
```
HerbertYiga, line 31: do we expect cross site scripting in such cases,ie on the input tag

omod/src/main/webapp/template/headerFull.jsp
```diff
@@ -178,7 +176,7 @@
 				<div id="openmrs_msg"><openmrs:message code="${msg}" text="${msg}" arguments="${msgArgs}"  htmlEscape="false" /></div>
 			</c:if>
 			<c:if test="${err != null}">
-				<div id="openmrs_error"><openmrs:message code="${err}" text="${err}" arguments="${errArgs}" htmlEscape="false"/></div>
+				<div id="openmrs_error"><openmrs:message code="${err}" text="${err}" arguments="${errArgs}" htmlEscape="true"/></div>
```
isears, line 179: @dkayiwa do you know why the htmlEscaping here was set to "false" originally? Can you think of anything in the legacy UI that would break by turning it on?

omod/src/main/webapp/admin/maintenance/implementationIdForm.jsp
```diff
@@ -18,7 +18,7 @@
 				<td style="white-space: nowrap">
 				    <input type="text" name="${status.expression}" value="${status.value}" size="40" />
 				    <c:if test="${status.errorMessage != ''}">
-                        <span class="error">${status.errorMessage}</span>
+                        <span class="error">$<c:out value="${status.errorMessage}"/></span>
```
isears, line 21: Was the `$` character before the `<c:out` tag on this line left in on purpose?

isears (commented):
@jmmacdo4 this is looking pretty good, I just want to take a little more time to make sure that turning on html filtering in `headerFull.jsp` doesn't break the UI elsewhere.

omod/src/main/webapp/admin/maintenance/implementationIdForm.jsp
jmmacdo4: No, thanks for pointing that out

omod/src/main/webapp/admin/maintenance/implementationIdForm.jsp
jmmacdo4: If the input field is loaded in with data from the server initially. So if a script is stored on the server, it could be executed when loaded. That was my thought process, but it could be incorrect. What do you think @isears?

isears: Are the patches in `implementationIdForm.jsp` still related to EMPT54? So far I've only been able to trigger XSS in the error messages that come up in `headerFull.jsp` but maybe I'm missing something.

isears (changes requested):

jmmacdo4:
@isears I have updated the sanitization to be present in the controller.

On this note, I think it would be really easy to CTRL-F `WebConstants.OPENMRS_ERROR_ATTR` and add these escape utils to knock out a bunch of XSS vulnerabilities. Is this something we want to look into or talk about more on Friday?

isears:
> On this note, I think it would be really easy to CTRL-F `WebConstants.OPENMRS_ERROR_ATTR` and add these escape utils to knock out a bunch of XSS vulnerabilities. Is this something we want to look into or talk about more on Friday?

@jmmacdo4 I'm warming up to the idea. It's possible that [the example](https://github.com/openmrs/openmrs-module-legacyui/blob/2042e094c48da0ff7aab20a0e0841dadeb09855e/omod/src/main/java/org/openmrs/web/controller/concept/ConceptDatatypeListController.java#L80) I cited above is the only thing you have to refactor.

If that's the case, what's your workaround for formatting that specific exception without using HTML tags?

jmmacdo4:
@isears I have went ahead with the error div "resign" we talked about. I have replaced all of the <br /> error messages with commas to allow us to turn `htmlEscape` to `true`. I tested a few of them out and it seems to be working okay.

isears:
Going to close this in favor of #149

jmmacdo4:
This PR has been restructured in #151

PR #139, opened 2021-03-09
Fixed XSS vulnerability in roles field of user list

annaowens:

@isears 
Fixed `users.jsp` to interpret the user roles (specifically when searching for users in User Management) as as text rather than as HTML. 

### Link to Ticket

### Issue I worked on

This fix protects against reflected XSS that is inputted into the title of a new user role. When the role is assigned to a user and that user is searched in the Manage Users tab of Advanced administration, the script is interpreted as HTML rather than text.  
  
_Note:_ Without the additional `c:out` tag on line 93 of `users.jsp`, a role with an XSS title will sometimes cause the `title` attribute to be vulnerable to XSS injection.

### Before Fix
Pop-ups with the message from the injected script tag:  
AddRole
  
User role is not shown in the UI as it is part of the HTML:   
ScriptInHTMLInAddRole

### After Fix
User role is now shown as text rather than interpreted as HTML:  
FixedAddRole  

#### Steps to reproduce

3. Select “System Administration”
4. Select “Advanced Administration”
5. Select “Manage Roles” 
6. Select “Add Role”
7. In the “Role” title input field, enter `<script>alert(1);</script>`
8. Click "Save Role"
9. Navigate to the "Manage Users" tab
10. Enter "admin" in the "Find User on Name" search field and click "Search".
11. Click on "admin" under "System Id" in the search results table to edit the admin user.
12. Under Roles, select the check mark next to <script>alert(1);</script>
13. Click "Save User"
14. Repeat step 10
  
_An alert with the text "1" will appear on the screen._

### DIFF
--- a/omod/src/main/webapp/admin/users/users.jsp
+++ b/omod/src/main/webapp/admin/users/users.jsp
@@ -90,7 +90,7 @@
 			<td><c:out value="${user.familyName}"/></td>
 			<td>
 				<c:if test="${fn:length(userRolesMap[user]) > 3}">
-				<span title="${userRolesMap[user]}">
+				<span title="<c:out value='${userRolesMap[user]}'/>">
 				</c:if>
 				<c:forEach var="r" items="${userRolesMap[user]}" varStatus="varStatus" end="2">
 					<c:choose>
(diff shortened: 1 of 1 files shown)


HerbertYiga (approved):

HerbertYiga:
thanks @annaowens

PR #147, opened 2021-03-25
RA-1875: EMPT55 Fixed XSS vulnerability in Manage Locales And Themes 

CrooksDC:
# Description of what I changed
Added c:out for the theme in the localesAndThemes file

# Issue I worked on
Sanitized input for when you add a new theme that is a script. The script would be executed when you try to submit the new theme. The script will not be executed with this patch.

It can be reproduced following

1. Provide following info and click login​Username​ as ​admin​ Password​ ​ as ​admin123 Location as ​Inpatient
2. Click on “System administration”
3. Then, click on “Advanced administration”
4. Click on “Manage Locales And Themes”
5. Enter the following details: Default Locale as en_GB Default theme as “><script>alert(12345)</script>
6. Click on “Submit”

# Link to ticket
https://issues.openmrs.org/browse/RA-1875

### DIFF
--- a/api/pom.xml
+++ b/api/pom.xml
@@ -15,7 +15,7 @@
 	<parent>
 		<groupId>org.openmrs.module</groupId>
 		<artifactId>legacyui</artifactId>
-		<version>1.8.2-SNAPSHOT</version>
+		<version>1.8.3-SNAPSHOT</version>
 	</parent>
 
 	<artifactId>legacyui-api</artifactId>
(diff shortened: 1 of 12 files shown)


HerbertYiga (approved):
Thanks @CrooksDC  cc @isears

isears (approved):
Thanks for the notification @HerbertYiga 

LGTM!

PR #138, opened 2021-03-08
EMPT-67: Fixed XSS vulnerability in alert header

annaowens:
### Description of what I changed

@isears 
Fixed the display format of alerts in `headerFull.jsp` to display the alert titles as text rather than as HTML. 

### Link to Ticket

https://issues.openmrs.org/browse/RA-1865  

### Issue I worked on

This fix protects against reflected XSS that is inputted to an alert title. When a script or other malicious input is inserted, it will be executed as HTML each time a page with the header is loaded for the user to which the alert was assigned. The page is loaded on nearly every legacy UI page. The fix is locaed in the `headerFull.jsp` file, as this file is the one that displays the alert title and is vulnerable to the XSS attack.

### Before Fix
Pop-ups with the message from the injected script tag:  
image  
  
Alert text is not shown as it is part of the HTML now:  
image
  

### After Fix
Alert title is now shown as text rather than interpreted as HTML
image

#### Steps to reproduce

1. Launch the OpenMRS application.
2. Login with username "Admin" and password "Admin123" with location as Inpatient Ward.
3. Select “System Administration”
4. Select “Advanced Administration”
5. Select “Manage Alerts” 
6. Select “Add Alert”
7. In the “Alert Text” input field, enter `<script>alert(1)</script>`
8. Next to the "Recipients" input field, click "Add", then select user "Super User" (_Note:_ it is important that you choose this user so that the alert appears for the currently logged in user, which is Super User since we logged in as Admin)
9. Click "Save Alert"
  
_An alert with the text "1" will appear on the screen each time a page with the alert header is loaded._

PR #151, opened 2021-03-27
Remove `<br />` tags from legacy ui error messages

jmmacdo4:
By remove `<br />` tags from legacy ui error messages, we can turn HTML escaping on and avoid the possibility of XSS attacks.

### DIFF
--- a/omod/src/main/java/org/openmrs/web/controller/maintenance/ImplementationIdFormController.java
+++ b/omod/src/main/java/org/openmrs/web/controller/maintenance/ImplementationIdFormController.java
@@ -23,6 +23,7 @@ import org.springframework.validation.BindException;
 import org.springframework.web.servlet.ModelAndView;
 import org.springframework.web.servlet.mvc.SimpleFormController;
 import org.springframework.web.servlet.view.RedirectView;
+import org.apache.commons.lang.StringEscapeUtils;
 
 /**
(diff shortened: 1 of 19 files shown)


jmmacdo4:
@isears Here is the `<br />` to comma rework PR.

omod/src/main/java/org/openmrs/web/controller/maintenance/ImplementationIdFormController.java
```diff
@@ -23,6 +23,7 @@
 import org.springframework.web.servlet.ModelAndView;
 import org.springframework.web.servlet.mvc.SimpleFormController;
 import org.springframework.web.servlet.view.RedirectView;
+import org.apache.commons.lang.StringEscapeUtils;
```
HerbertYiga, line 26: do we need this change?

jmmacdo4: No, thanks Herbert I fixed that.

HerbertYiga (approved):
looks good to me  cc @isears

isears (approved):
Ok, this looks good. Thanks @jmmacdo4 and thanks for reviewing @HerbertYiga !

Also looks like the merge conflict with #138 can be resolve automatically

PR #137, opened 2021-03-03
EMPT-48: Fixed XSS vulnerability in Manage Fields list

annaowens:
### Description of what I changed

@annaowens
Fixed the HTML parsing in `OpenmrsSearch.js` to read users' search input as text rather than HTML. 

_Note_: I did not fight many documented fixes of vulnerabilities in `.js` files, so I modeled this fix after [this PR](https://github.com/openmrs/openmrs-module-legacyui/pull/87/files). Please let me know if a different technique to fix the vulnerability is preferred. This is also my first PR to OpenMRS, so let me know if if you have any suggestions to improve the PR format in general. Thanks!

### Issue I worked on

This fix protects against reflected XSS in the search results on the Admin's Manage Fields page (in Advanced Administration).

**Before Fix**
image

**After Fix**
image

#### Steps to reproduce

1. Launch OpenMRS application.
2. Login with username "Admin" and password "Admin123" with location as Inpatient Ward.
3. Select “System Administration”
4. Select “Advanced Administration”
5. Select “Manage Fields”
6. In the “Find Fields by Name” input field, enter `<iframe src="https://www.ncsu.edu"></iframe>`

PR #152, opened 2021-03-30
RA-1865 Added HTML escape for search results

maxhark:
When building the search results table, I changed `return data` to ` return $j('<div/>').text(data).html()` in order to prevent Iframes from being shown in the person search results.

This is a very similar issue to the one fixed by Anna [here](https://github.com/openmrs/openmrs-module-legacyui/pull/137).

### Link to ticket

### Issue I worked on
One could create a new person with given name `<iframe src="http://www.ncsu.edu">`. When searching for this person by typing `<iframe`, the results list would display that iframe. In addition, where it says "viewing results for __" would also display a blank iframe.

### Before fix
Searching for `<iframe` when a person with given name `<iframe src="http://www.ncsu.edu">` exists.
image

### After fix
Searching for `<iframe` when a person with given name `<iframe src="http://www.ncsu.edu">` exists.
image

### Steps to reproduce
1. Login to OpenMRS as admin.
2. Go to System Admininstration > Advanced Administration > Manage Persons > Create Person
3. Create a person with name `<iframe src="http://www.ncsu.edu">`
4. On the page with advanced details make sure that the given name is `<iframe src="http://www.ncsu.edu">`, as it will automatically split into given name and last name otherwise.
5. Save this person.
6. Go back to Manage Persons.
7. Search for `<iframe`
8. The iframe of the site will be displayed in the results, and a blank iframe will be shown after the "viewing results for.." text.

### DIFF
--- a/omod/src/main/webapp/resources/scripts/jquery-ui/js/openmrsSearch.js
+++ b/omod/src/main/webapp/resources/scripts/jquery-ui/js/openmrsSearch.js
@@ -841,8 +841,7 @@ function OpenmrsSearch(div, showIncludeVoided, searchHandler, selectionHandler, 
                 var data = rowData[c.fieldName];
                 if(data == null)
                     data = " ";
-
-                return data;
+                return $j('<div/>').text(data).html();
             });
 
             //include the attributes
--- a/omod/src/main/webapp/resources/scripts/jquery-ui/js/openmrsSearch.js
+++ b/omod/src/main/webapp/resources/scripts/jquery-ui/js/openmrsSearch.js
@@ -823,7 +823,7 @@ function OpenmrsSearch(div, showIncludeVoided, searchHandler, selectionHandler, 
                 $j('#openmrsSearchTable_paginate').show();
             }
 
-            this._updatePageInfo(searchText);
+            this._updatePageInfo($j('<div/>').text(searchText).html());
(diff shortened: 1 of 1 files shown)


omod/src/main/webapp/resources/scripts/jquery-ui/js/openmrsSearch.js
```diff
@@ -841,8 +841,7 @@ function OpenmrsSearch(div, showIncludeVoided, searchHandler, selectionHandler,
                 var data = rowData[c.fieldName];
                 if(data == null)
                     data = " ";
-
-                return data;
+                return $j('<div/>').text(data).html();
```
HerbertYiga, line 844: cc @isears

HerbertYiga (approved):

maxhark:
Added sanitization of one more field, in some situations typing `<iframe` was still displaying an iframe as part of the "viewing results for..." text. This change should fix that. 

For what its worth, this fixes EMPT-46 as well, which uses `openMrsSearch.js` the same way, just for Reference Terms instead of Persons.

isears (approved):
I'm generally not a huge fan of patching in the .js, but we've done it before and maybe this is the best option if it fixes multiple.

Thanks @maxhark !

PR #153, opened 2021-03-30
RA-1865: Fix XSS Vulnerability in Manage Fields List

annaowens:
### Description of what I changed

@isears 
Fixed `DWRFormService.java` to encode elements of the `FieldListItem` object before passing it to the UI to display in the search results on the `Manage Fields` page.

_Note_: Since `ui` was not defined in this file, I modeled this fix after [this PR](https://github.com/openmrs/openmrs-module-uiframework/pull/52/files).

### Link to Ticket

https://issues.openmrs.org/browse/RA-1865  

### Issue I worked on

This fix protects against stored XSS that is inputted into the `Name` when creating a new Field. After creating the field, when you search that field in the Manage Fields page, the search result table will execute the XSS vulnerability in the results.

### Before Fix
Injected iframe appears in search results:
VulnerableEMPT49

### After Fix
Field name is now shown as text rather than interpreted as HTML:   
FixedEMPT49

#### Steps to reproduce

1. Launch the OpenMRS application.
2. Login with username "Admin" and password "Admin123" with location as Inpatient Ward.
3. Select “System Administration”
4. Select “Advanced Administration”
5. Select “Manage Fields” 
6. Select “Add New Field”
7. In the “Field Name” input field, enter `<iframe src=https://www.csc.ncsu.edu/>`
8. Click "Save"
9. Navigate to the "Manage Fields" tab
10. In the search bar, type "iframe". Do not click enter, the search results will populate.

_An iframe will appear in the search result row of the encounter you created under the "Name" column._

PR #155, opened 2021-04-07
RA-1865: Fix XSS vulnerability in Field Type and Description of Manage Fields

annaowens:

@isears 
Fixed `DWRFormService.java` to encode the `fieldType` and `description` attributes of the `FieldListItem` object before passing it to the UI to display in the search results on the `Manage Fields` page. 

**This change is very similar to [this PR](https://github.com/openmrs/openmrs-module-legacyui/pull/153) and probably should have been fixed there had I realized this were a vulnerability. I went ahead and tested the Description column for a vulnerability as well, and it had one that I fixed (again, should have been fixed in the first PR).

_Note_: Since `ui` was not defined in this file, I modeled this fix after [this PR](https://github.com/openmrs/openmrs-module-uiframework/pull/52/files).

### Link to Ticket

### Issue I worked on

This fix protects against stored XSS that is inputted into the name of a new `Field Type` object that is later assigned to a new Field and the `Description` attribute of a new Field. After creating the field, when you search that field in the Manage Fields page, the search result table will execute the XSS vulnerability in the results under the `Field Type` and `Description` columns.

### Before Fix
Injected iframe appears in search results:

Field Type:   
VulnerableEMPT50  

Description:  
VulnerableEMPT50-Part2

### After Fix
Field type is now shown as text rather than interpreted as HTML:   
FixedEMPT50

Description is now shown as text rather than interpreted as HTML:   
FixedEMPT50-Part2  

#### Steps to reproduce

3. Select “System Administration”
4. Select “Advanced Administration”
1. Click on “Manage Field Types” and then “Add Field Type”.
1. Enter `<iframe src=’http://ncsu.edu’>` as Name.
1. Hit Save Field Type.
1. Click on “Manage Fields” and then “Add New Field”.
1. Enter “test” in the Field Name text box and select <iframe src=’http://ncsu.edu’> for Field Type.
1. Enter `<iframe src=’http://ncsu.edu’>` as Description.
1. Hit Save.
1. In the Find Fields By Name text box, enter “test”. Do not click enter, the results will populate automatically.

_An iframe will appear in the search result row of the field you created under the "Field Type" and "Description" columns._

### DIFF
--- a/omod/src/main/java/org/openmrs/web/dwr/DWRFormService.java
+++ b/omod/src/main/java/org/openmrs/web/dwr/DWRFormService.java
@@ -109,6 +109,14 @@ public class DWRFormService {
 		for (Field field : Context.getFormService().getFields(txt)) {
 			FieldListItem htmlSafeField = new FieldListItem(field, Context.getLocale());
 			htmlSafeField.setName(Encode.forHtml(htmlSafeField.getName()));
+			htmlSafeField.setFieldTypeName(Encode.forHtml(htmlSafeField.getFieldTypeName()));
+			
+			// Add the description only if it was given
+			String description = htmlSafeField.getDescription();
+			if (description != null) {
+				htmlSafeField.setDescription(Encode.forHtml(description));
+			}
+			
 			fields.add(htmlSafeField);
 		}
 		


omod/src/main/java/org/openmrs/web/dwr/DWRFormService.java
```diff
@@ -109,6 +109,14 @@ public FormFieldListItem getFormField(Integer formFieldId) {
 		for (Field field : Context.getFormService().getFields(txt)) {
 			FieldListItem htmlSafeField = new FieldListItem(field, Context.getLocale());
 			htmlSafeField.setName(Encode.forHtml(htmlSafeField.getName()));
+			htmlSafeField.setFieldTypeName(Encode.forHtml(htmlSafeField.getFieldTypeName()));
+			
+			// Add the description only if it was given
+			String description = htmlSafeField.getDescription();
```
annaowens, line 115: The reason for this check was that if a description was not given for a new field (the field is optional), "null" would appear for that row under the `Description` column which looked bad. I guess this was a product of the encoding, but the `if` condition checking that the description was not null before encoding it fixed the issue, allowing for blank values in the `Description` column rather than many "null"s.

omod/src/main/java/org/openmrs/web/dwr/DWRFormService.java
```diff
@@ -109,6 +109,14 @@ public FormFieldListItem getFormField(Integer formFieldId) {
 		for (Field field : Context.getFormService().getFields(txt)) {
 			FieldListItem htmlSafeField = new FieldListItem(field, Context.getLocale());
 			htmlSafeField.setName(Encode.forHtml(htmlSafeField.getName()));
+			htmlSafeField.setFieldTypeName(Encode.forHtml(htmlSafeField.getFieldTypeName()));
+
```
HerbertYiga, line 113: @annaowens  you can remove this space

isears: I think the line break enhances readability a bit by separating the name encoding from the description encoding. It doesn't look like it violates any [OpenMRS coding conventions](https://wiki.openmrs.org/display/docs/Java+Conventions#JavaConventions-CodeStyle), either.

isears (approved):
Thanks for digging a little deeper on the null descriptions @annaowens

PR #156, opened 2021-04-07
Logging when user accesses server log

jmmacdo4:
@isears one of our listed vulnerabilities was that no logging occurs when a user views the server log. This should be a simple fix.

### DIFF
--- a/omod/src/main/java/org/openmrs/web/controller/maintenance/ServerLogController.java
+++ b/omod/src/main/java/org/openmrs/web/controller/maintenance/ServerLogController.java
@@ -76,6 +76,7 @@ public class ServerLogController extends SimpleFormController {
 				}
 				
 				MemoryAppender memoryAppender = (MemoryAppender) getMemoryAppenderHandle.invoke();
+				log.info("Server log was accessed");
 				return memoryAppender.getLogLines();
 			}
 			catch (Throwable e) {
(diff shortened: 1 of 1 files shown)


tejasdeshpande100 (approved):

harshit151001 (approved):
Looks good to me!

HerbertYiga:
> one of our listed vulnerabilities was that no logging occurs when a user views the server log. This should be a simple fix.

@jmmacdo4  did this have a ticket Id?

omod/src/main/java/org/openmrs/web/controller/maintenance/ServerLogController.java
```diff
@@ -76,6 +76,7 @@ protected ModelAndView onSubmit(HttpServletRequest request, HttpServletResponse
 				}
 				
 				MemoryAppender memoryAppender = (MemoryAppender) getMemoryAppenderHandle.invoke();
+				log.info("Server log was accessed");
```
HerbertYiga, line 79: is there a better method of having this than simply hard coding stuff,ie under https://github.com/openmrs/openmrs-module-legacyui/blob/master/api/src/main/resources/messages.properties ,hi  @isears what do you think?

isears: @HerbertYiga that's a fair point, but it seems like most existing calls to `log.info()` use hard-coded strings rather than message properties ([examples](https://github.com/search?p=1&q=org%3Aopenmrs+log.info&type=Code)).

isears (commented):
@jmmacdo4 it seems like logging would be more appropriate closer to the actual DB access. For example, if the logs are accessed from a module outside the legacy-ui, it doesn't seem like this change would catch that access.

Have you considered logging somewhere in openmrs-core to catch log-reads from arbitrary modules?

HerbertYiga:
@isears  thanks isaac

isears (approved):
As we discussed at the sync meeting today, logs aren't actually written to the db and this may actually be the best way to catch log-reads. My mistake.

Thanks @jmmacdo4 !

PR #158, opened 2021-04-14
Changed DWRPersonService for EMPT60

Parth59:
Link to ticket

Reverted the changes for https://github.com/openmrs/openmrs-module-legacyui/pull/152 and add a serverside fix.

The issue I worked on
One could create a new person with given name <iframe src="http://www.ncsu.edu">. When searching for this person by typing <iframe, the results list would display that iframe. In addition, where it says "viewing results for __" would also display a blank iframe.

Before fix
Searching for <iframe when a person with given name <iframe src="http://www.ncsu.edu"> exists.
image

After fix
Searching for <iframe when a person with given name <iframe src="http://www.ncsu.edu"> does not exists.
image

Steps to reproduce
Login to OpenMRS as admin.
Go to System Admininstration > Advanced Administration > Manage Persons > Create Person
Create a person with name <iframe src="http://www.ncsu.edu">
On the page with advanced details make sure that the given name is <iframe src="http://www.ncsu.edu">, as it will automatically split into given name and last name otherwise.
Save this person.
Go back to Manage Persons.
Search for <iframe
The iframe of the site will be displayed in the results, and a blank iframe will be shown after the "viewing results for.." text.

### DIFF
--- a/omod/src/main/java/org/openmrs/web/dwr/DWRPersonService.java
+++ b/omod/src/main/java/org/openmrs/web/dwr/DWRPersonService.java
@@ -32,6 +32,7 @@ import org.openmrs.api.PatientService;
 import org.openmrs.api.PersonService;
 import org.openmrs.api.UserService;
 import org.openmrs.api.context.Context;
+import org.openmrs.web.WebUtil;
 
 /**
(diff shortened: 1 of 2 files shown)


isears (approved):
Looks great thanks @Parth59

PR #160, opened 2021-04-16
RA-1875: EMPT66 Fixed XSS Vulnerability in Schedule Form

The-Lady:
### Description of What I Changed

I added encoding to SchedulerForm java file to prevent XSS vulnerability in Task Name.

### Issue I Worked On

An iframe/script could be injected in Task Name while adding a new task in Scheduler. I encoded the name of the task in the controller file to mitigate this.

Steps to reproduce the vulnerability:

3. Click on the “System Administration” option and then on "Advanced Administration".
4. Click on “Manage Scheduler”.
5. Click on “Add Tasks”.
6. Enter '<iframe src="https://www.csc.ncsu.edu/">' as Name and “org.openmrs.scheduler.tasks.SendEmailTask” and Schedulable Class.
7. Click on “Save”.

Output: An iframe will be displayed.

### Link to ticket
[RA-1875](https://issues.openmrs.org/browse/RA-1875)

@isears

### DIFF
--- a/omod/src/main/java/org/openmrs/web/dwr/DWRConceptService.java
+++ b/omod/src/main/java/org/openmrs/web/dwr/DWRConceptService.java
@@ -222,8 +222,10 @@ public class DWRConceptService {
 			
 			if (searchResults.size() < 1) {
 				String htmlSafePhrase = "<b>" + Encode.forHtml(phrase) + "</b>";
-				objectList.add(Context.getMessageSourceService().getMessage("general.noMatchesFoundInLocale",
-				    new Object[] { htmlSafePhrase, OpenmrsUtil.join(searchLocales, ", ") }, Context.getLocale()));
+				objectList.add(Context.getMessageSourceService()
+				        .getMessage("general.noMatchesFoundInLocale",
+				            new Object[] { htmlSafePhrase, OpenmrsUtil.join(searchLocales, ", ") },
+				            Context.getLocale()));
 			} else {
 				// turn searchResults into concept list items
 				// if user wants drug concepts included, append those
--- a/omod/src/main/java/org/openmrs/web/dwr/DWRConceptService.java
+++ b/omod/src/main/java/org/openmrs/web/dwr/DWRConceptService.java
@@ -222,8 +222,10 @@ public class DWRConceptService {
 			
 			if (searchResults.size() < 1) {
 				String htmlSafePhrase = "<b>" + Encode.forHtml(phrase) + "</b>";
-				objectList.add(Context.getMessageSourceService().getMessage("general.noMatchesFoundInLocale",
-				    new Object[] { htmlSafePhrase, OpenmrsUtil.join(searchLocales, ", ") }, Context.getLocale()));
+				objectList.add(Context.getMessageSourceService()
+						.getMessage("general.noMatchesFoundInLocale",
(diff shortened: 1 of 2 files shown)


omod/src/main/java/org/openmrs/web/dwr/DWRConceptService.java
```diff
@@ -222,10 +222,8 @@
 			
 			if (searchResults.size() < 1) {
 				String htmlSafePhrase = "<b>" + Encode.forHtml(phrase) + "</b>";
-				objectList.add(Context.getMessageSourceService()
-				        .getMessage("general.noMatchesFoundInLocale",
-				            new Object[] { htmlSafePhrase, OpenmrsUtil.join(searchLocales, ", ") },
-				            Context.getLocale()));
+				objectList.add(Context.getMessageSourceService().getMessage("general.noMatchesFoundInLocale",
```
HerbertYiga, line 225: @The-Lady  did you make any code changes around this place?

isears (changes requested):
@The-Lady this looks good but it seems like the changes to `DWRConceptService.java` aren't necessary (as @HerbertYiga pointed out).

If that's the case I'll go ahead and merge this in after the `DWRConceptService.java` changes are reset.

omod/src/main/java/org/openmrs/web/dwr/DWRConceptService.java
The-Lady: @HerbertYiga I am sorry, I did not notice the unintentional change in indentation. I have reverted the change and committed to the branch.

The-Lady:
@isears I have reset the changes in `DWRConceptService.java`

Parth59:
Hi @The-Lady .  Can we try to import the org.openmrs.web.WebUtil package to utilize WebUtil.escapeHTML() function () ? @isears Will it be a better idea ?

isears (changes requested):
There's still changes in DWRConceptService.java as part of this PR. Just revert the file to the original so that we can keep the history clean. Something like this should do the trick: 
```
git checkout origin/master omod/src/main/java/org/openmrs/web/dwr/DWRConceptService.java 
```
> Hi @The-Lady . Can we try to import the org.openmrs.web.WebUtil package to utilize WebUtil.escapeHTML() function () ? @isears Will it be a better idea ?

Yes, in general the webutil escaping functions (which are backed by OWASP) are more robust than apache string escape utils. If it's possible, let's use webutil, but if it creates issues don't worry to much about it.

The-Lady:
@Parth59 Thank you for the suggestion

@isears I have reverted the DWRConcept file to original and switched to WebUtil for encoding as suggested.

isears:
@The-Lady thanks for switching to webUtil, that looks good.

It looks like `DWRConceptService.java` is still showing up as a change in the git diff of this PR. Were you able to run the git command I posted above? The end result should be that only `SchedulerFormController.java` shows up in the "Files changed" tab.

The reason it's so important to not include modification to files that don't have to be modified as part of a patch is so that we can more easily trace the origins of bugs that get introduced into the codebase.

The-Lady:
@isears The indentation kept getting changed whenever I run the application on localhost. I have reverted it again.

isears (approved):
Great, thanks @The-Lady !

PR #164, opened 2021-06-07
Added Changes to UserFormController.java to redirect error message back to user.

Parth59:
Hi Team, @isears 

This change is for fixing EMPT70. I have performed minor test runs to the best of my abilities. Additionally, there are minor one line addition to message.properties file in openmrs-core. I will shortly raise them and link this PR in those comments.

Steps to reproduce :
i) Go to Advance Administration ‹ Manage Users
ii) Enter admin in find user on name text field
iii)Click Search
iv)Click admin in the search result
v)Enter javascript:/*--&gt;&lt;/title&gt;&lt;/style&gt;&lt;/textarea&gt;&lt;/script&gt;&lt;/xmp&gt;&lt;svg/onlo ad='+/"/+/onmouseover=1/+/[*/[]/+alert(1)//'&gt; in the first name, middle name and last name input fields
vi) Click Save

Before fix:

image

After fix :
EMPT70


omod/src/main/java/org/openmrs/web/controller/user/UserFormController.java
```diff
@@ -270,9 +271,20 @@ public String handleSubmission(WebRequest request, HttpSession httpSession, Mode
...
-			
 			userValidator.validate(user, errors);
-			
+			if (errors.hasErrors()) {
+				List<FieldError> fieldErrorList = errors.getFieldErrors();
+				for (FieldError fieldError : fieldErrorList) {
+					String[] errorCodes = fieldError.getCodes();
+					for (String value : errorCodes) {
+						if (value.contains("error.exceededMaxLengthOfField")) {
+							httpSession.setAttribute(WebConstants.OPENMRS_ERROR_ATTR, "User.DemographicInfo.length");
```
HerbertYiga, line 281: thanks @Parth59 ,some times sessions get disturbing more so when they are not destroyed after use,why not use model .addAttribute( )

isears (commented):
Thanks @Parth59 

If this PR depends on https://github.com/Parth59/openmrs-core/pull/1, then we will have to add the next release of openmrs-core as a dependency.

I wonder would it be possible to just add your message here (https://github.com/openmrs/openmrs-module-legacyui/blob/5f8d1d3ea5844cedb535b1634d9848ba0ca3b320/api/src/main/resources/messages.properties) so that this change doesn't depend on updates to external modules?

omod/src/main/java/org/openmrs/web/controller/user/UserFormController.java
Parth59: Hi @HerbertYiga , I have started reading about the same. Will try to implement those changes in the code !

Parth59:
> Thanks @Parth59
> 
> If this PR depends on [Parth59/openmrs-core#1](https://github.com/Parth59/openmrs-core/pull/1), then we will have to add the next release of openmrs-core as a dependency.
> 
> I wonder would it be possible to just add your message here (https://github.com/openmrs/openmrs-module-legacyui/blob/5f8d1d3ea5844cedb535b1634d9848ba0ca3b320/api/src/main/resources/messages.properties) so that this change doesn't depend on updates to external modules?

Sure. I will test it out if this approach works.

Parth59:
Hi Issac,
The approach you suggested worked ! Requesting you to kindly review the changes.

omod/src/main/java/org/openmrs/web/controller/user/UserFormController.java
Parth59: Hi @HerbertYiga

I tried using the model.addAttrribute() and it seems like using those with redirect causes them to be added to the URL. Additionally, the <c:if> tag in the view doesn't load up the error messages as can be seen in the screenshot referenced below.  [isLengthExceeded in the url comes to be true but isn't loaded in the page]

Infact, I also double checked for places that utilize model.addAttribute() in the same function. Turns out that the entire function utilizes httpSession for passing response back to the user. 
So requesting you to kindly let this security PR pass . If needed I raise an issue and then will refactor to modify the entire function in a separate PR. Sounds good ? 

ScreenShot

omod/src/main/java/org/openmrs/web/controller/user/UserFormController.java
```diff
@@ -273,6 +274,19 @@ public String handleSubmission(WebRequest request, HttpSession httpSession, Mode
 			
 			userValidator.validate(user, errors);
 			
+			if (errors.hasErrors()) {
```
isears, line 277: Can we combine the logic of this if-clause with the if-clause immediately below it (line 290)?

It looks like the condition is the same and collapsing them into a single if-clause might make the error-handling logic of this section a little more readable.

isears (changes requested):
Thanks @Parth59

The result is good, just needs a bit of refactoring to make the code cleaner.

omod/src/main/java/org/openmrs/web/controller/user/UserFormController.java
Parth59: Hi Issac, I have merged in the code as per suggestions.

omod/src/main/java/org/openmrs/web/controller/user/UserFormController.java
```diff
@@ -273,7 +274,23 @@ public String handleSubmission(WebRequest request, HttpSession httpSession, Mode
...
 			if (errors.hasErrors()) {
+
+				//check if length exceeded in DemographicInfo fields
+				List<FieldError> fieldErrorList = errors.getFieldErrors();
+				for (FieldError fieldError : fieldErrorList) {
+					String[] errorCodes = fieldError.getCodes();
+					for (String value : errorCodes) {
+						if (value.contains("error.exceededMaxLengthOfField")) {
+							httpSession.setAttribute(WebConstants.OPENMRS_ERROR_ATTR,"legacyui.manageuser.DemographicInfo.lengthExceeded");
+							httpSession.setAttribute(WebConstants.OPENMRS_ERROR_ARGS, "50");
```
sherrif10, line 287: Hello @Parth59 . How far with this ticket.

PR #169, opened 2021-07-30
MM-893: Solved Authentication Bypass to System Administration

jnsereko:
A user without the **
Uploading 69ca3ccd-5a45-4d3b-bc40-cbcc4f576dc5.mp4…
** privilege could access the administration page and all other functionalities.  
I have just created a Filter to redirect every user, without the above privilege, to the home page.

cc @isears @sherrif10


omod/src/main/java/org/openmrs/module/web/filter/AdminAuthorisationFilter.java
```diff
@@ -0,0 +1,65 @@
...
+ * This filter checks if an authenticated user trying to access administrative functions has the
+ * <STRONG>System Administration<STRONG> privelege. It will intercept any requests with *admin/* in
+ * its url. Unauthorised user will be redirected to the home page.
+ */
+public class AdminAuthorisationFilter implements Filter {
+	
+	/**
+	 * @see Filter#init(FilterConfig)
+	 */
+	public void init(FilterConfig filterConfig) throws ServletException {
```
sherrif10, line 37: What does this method do ?

jnsereko: @isears, I think It generally does nothing other than obeying java laws.  
As far as I know, javax.servelt.Filter class is an interface so when implementing it, I have to implement [all its methods ](https://javaee.github.io/javaee-spec/javadocs/javax/servlet/Filter.html )otherwise the java-compiler will complain

isears (changes requested):
This mostly looks good to me, just had a few small adjustments.

@dkayiwa would you mind taking a look at this as well? This is a solution to the problem mentioned here (https://talk.openmrs.org/t/proposal-for-securing-the-legacy-ui-system-administration-pages/33988) although we took a slightly different approach to avoid having to edit every single admin page.

omod/src/main/java/org/openmrs/module/web/filter/AdminAuthorisationFilter.java
```diff
@@ -0,0 +1,65 @@
...
+	/**
+	 * @see javax.servlet.Filter#doFilter(javax.servlet.ServletRequest,
+	 *      javax.servlet.ServletResponse, javax.servlet.FilterChain)
+	 */
+	public void doFilter(ServletRequest req, ServletResponse res, FilterChain chain) throws IOException, ServletException {
+		HttpServletRequest httpReq = (HttpServletRequest) req;
+		String requestURI = httpReq.getRequestURI();
+		if (requestURI.contains("/admin/")) {
+			User user = Context.getAuthenticatedUser();
+			if (user != null && !user.hasPrivilege("App: coreapps.systemAdministration")) {
```
dkayiwa, line 50: How about those using the legacyui module without the reference application modules?

isears: Ah! I knew we were missing something! Could we instead implement this filter in the refapp module?

omod/src/main/java/org/openmrs/module/web/filter/AdminAuthorisationFilter.java
```diff
@@ -0,0 +1,65 @@
...
+				httpReq.getSession().setAttribute(WebConstants.DENIED_PAGE, requestURI);
+				HttpServletResponse httpRes = (HttpServletResponse) res;
+				httpRes.sendRedirect(httpReq.getContextPath() + "/login.htm");
```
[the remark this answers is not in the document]
jnsereko: Hey @isears! Thanks for the review 
I definitely agree that is so important for us to log a "missing privilege message" on page denial.

omod/src/main/java/org/openmrs/module/web/filter/AdminAuthorisationFilter.java
jnsereko: Thank @isears and @dkayiwa for the rewie.  I have really been thinking a lot about this.  
what if we try something like
 ```
if (!(user.isSuperUser() || user.hasRole("Organizational: System Administrator") || user.hasRole("Organizational: Hospital Administrator"))) 
```
Will it also depend on the reference application

**Comment by jnsereko** [comment]
https://user-images.githubusercontent.com/58003327/127879316-8c1b5bb9-c16c-41d0-9129-bda5bb8ec2a6.mp4
Everything seams to be working fine

File: omod/src/main/java/org/openmrs/module/web/filter/AdminAuthorisationFilter.java [hunk 4]
```diff
@@ -0,0 +1,68 @@
...
+	/**
+	 * @see javax.servlet.Filter#doFilter(javax.servlet.ServletRequest,
+	 *      javax.servlet.ServletResponse, javax.servlet.FilterChain)
+	 */
+	public void doFilter(ServletRequest req, ServletResponse res, FilterChain chain) throws IOException, ServletException {
+		HttpServletRequest httpReq = (HttpServletRequest) req;
+		User user = Context.getAuthenticatedUser();
+
+		if (!(user.isSuperUser() || user.hasRole("Organizational: System Administrator") || user.hasRole("Organizational: Hospital Administrator"))) {
+			httpReq.getSession().setAttribute(WebConstants.DENIED_PAGE, httpReq.getRequestURI());
```
isears, line 54: I think we have to revert to checking for only `App: coreapps.systemAdministration`, like you had before.

To resolve the issue that @dkayiwa mentioned, let's just check for the existence of the `App: coreapps.systemAdministration` privilege before we do anything else. So there will be three cases:

1. The `App: coreapps.systemAdministration` privilege exists, and the user has it. Therefore, the user should be allowed to view the /admin/* pages
2. The `App: coreapps.systemAdministration` privilege exists, and the user does not have it. Therefore, the user should be prevented from viewing the /admin/* pages
3. The `App: coreapps.systemAdministration` does not exist, because the implementation is legacy-ui only. Therefore, we should let the user view the /admin/* pages to maintain backward compatibility.

isears (changes requested):
@jnsereko let me know if any of the above is unclear. If necessary we can set up a call to talk things through.

omod/src/main/java/org/openmrs/module/web/filter/AdminAuthorisationFilter.java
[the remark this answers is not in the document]
isears: Great, thanks for adding logging!

omod/src/main/java/org/openmrs/module/web/filter/AdminAuthorisationFilter.java
jnsereko: This looks very clear to me @isears and am sure it will work for us.  I need to come up with a way of checking if the system has the `App: coreapps.systemAdministration ` privilege

omod/src/main/java/org/openmrs/module/web/filter/AdminAuthorisationFilter.java
```diff
@@ -0,0 +1,78 @@
...
+	}
+	
+	/**
+	 * @see javax.servlet.Filter#doFilter(javax.servlet.ServletRequest,
+	 *      javax.servlet.ServletResponse, javax.servlet.FilterChain)
+	 */
+	public void doFilter(ServletRequest req, ServletResponse res, FilterChain chain) throws IOException, ServletException {
+		HttpServletRequest httpReq = (HttpServletRequest) req;
+		Collection<Module> modules = ModuleFactory.getLoadedModules();
+		if (modules.contains(ModuleFactory.getModuleById(COREAPPS_MODULE_ID))) {
```
jnsereko, line 60: hello @isears @sherrif10 @dkayiwa 

@dkayiwa, won't this line cater for those using the legacyui without reference application modules.

jnsereko:
It is still working fine 

https://user-images.githubusercontent.com/58003327/128019350-8b4f1e58-cb2a-49db-804f-e4d54bb42ab5.mp4

omod/src/main/java/org/openmrs/module/web/filter/AdminAuthorisationFilter.java
```diff
@@ -0,0 +1,78 @@
...
+ * <STRONG>System Administration<STRONG> privelege. It will intercept any requests with *admin/* in
+ * its url. Unauthorised user will be redirected to the home page.
+ */
+public class AdminAuthorisationFilter implements Filter {
+	
+	private static final Logger log = LoggerFactory.getLogger(AdminAuthorisationFilter.class);
+	
+	private static final String COREAPPS_MODULE_ID = "coreapps";
+	
+	private static final String COREAPPS_SYSTEM_ADMINISTRATOR_PRIVELEGE = "App: coreapps.systemAdministration";
```
dkayiwa, line 44: I wish the privilege did not have `coreapps` in its name because the legacyui module should not be aware of any module.

omod/src/main/java/org/openmrs/module/web/filter/AdminAuthorisationFilter.java
dkayiwa: Of course it would. But it is not a good practise for the legacyui module to be aware of any other modules.

PR #170, opened 2021-08-12
MM-918: Close all user sessions after password change and on logout

jnsereko:
Hello Everyone, I have tried to improve session management by keeping track of sessions opened by every user before. 
When a user changes his password or logs out, all sessions must be invalidated so that an attacker is trapped out.

As @ibacher had advised me, we had to even implement this feature in core but this was almost impossible because the classes I created weren't accessible in the adminui-module.
Thank you 
cc @isears @sherrif10 @ibacher 
https://user-images.githubusercontent.com/58003327/129126775-012991f1-4bae-4bea-8114-f76a754b17c7.mp4

### DIFF
--- a/omod/src/main/webapp/template/headerFull.jsp
+++ b/omod/src/main/webapp/template/headerFull.jsp
@@ -173,7 +173,7 @@
 			</openmrs:forEachAlert>
 
 			<c:if test="${msg != null}">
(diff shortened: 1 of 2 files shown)


jnsereko:
slaljfjkljfklskdf

PR #175, opened 2021-12-22
LUI-182, close active programs when exiting from care

cioan:
Hi @dkayiwa and @mogoodrich , here is the small change I made in the legacyui to bring back this feature. Thanks!

### DIFF
--- a/omod/src/main/java/org/openmrs/web/dwr/DWRPatientService.java
+++ b/omod/src/main/java/org/openmrs/web/dwr/DWRPatientService.java
@@ -570,18 +570,16 @@ public class DWRPatientService implements GlobalPropertyListener {
 							ret = "Internal error while trying to process patient death - unable to proceed. Cause: "
 							        + e.getMessage();
 						}
-						boolean closePrograms = Boolean.parseBoolean(Context.getAdministrationService().getGlobalProperty(
-						    "legacyui.closeProgramsWhenExitingFromCare", "false"));
-						if (closePrograms) {
-							try {
-								Context.getService(LegacyUIService.class).exitFromCare(patient, exitDate, exitReasonConcept);
-							}
-							catch (Exception e) {
-								log.warn("Caught error", e);
-								ret = "Internal error while trying to exit patient from care - unable to exit patient from care at this time. Cause: "
-								        + e.getMessage();
-							}
+						
+						try {
+							Context.getService(LegacyUIService.class).exitFromCare(patient, exitDate, exitReasonConcept);
 						}
+						catch (Exception e) {
+							log.warn("Caught error", e);
+							ret = "Internal error while trying to exit patient from care - unable to exit patient from care at this time. Cause: "
+							        + e.getMessage();
+						}
+						
 					}
 					// cause of death concept does not exist
 					else {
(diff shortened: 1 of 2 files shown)


omod/src/main/java/org/openmrs/web/dwr/DWRPatientService.java
```diff
@@ -570,6 +570,18 @@ public String exitPatientFromCare(Integer patientId, Integer exitReasonId, Strin
 							ret = "Internal error while trying to process patient death - unable to proceed. Cause: "
 							        + e.getMessage();
 						}
+						boolean closePrograms = Boolean.parseBoolean(Context.getAdministrationService().getGlobalProperty(
```
mogoodrich, line 573: Looking at the full "exitPatientFromCare" method it looks like this method ends up calling the "processDeath" method on a Patient if a death concept has been specified, and the reason for exit = death.  Otherwise, it calls the "exitFromCare" method.

I think what you were explaining to me yesterday (but I probably didn't follow fully) was that the "processDeath" method no longer closes programs?  

If I'm following this fully, I think actually we don't need to have a global property here and just call "exitFromCare" after "processDeath"... if right now calling exit from care with reason = Death does *not* exit them from care (but just makes them as dead) while calling the function with any other reason *does* exit from care, this seems like a bug. 

Thoughts @dkayiwa ?

mogoodrich (commented):
One question... after looking at the code and understanding better, we might be able to even simplify this further.

omod/src/main/java/org/openmrs/web/dwr/DWRPatientService.java
dkayiwa: According to the documentation, `processDeath` is supposed to exist the patient from care: https://github.com/openmrs/openmrs-core/blob/master/api/src/main/java/org/openmrs/api/PatientService.java#L568-L569

cioan: Thanks @dkayiwa ! Yes, processDeath calls [exitFromCare](https://github.com/openmrs/openmrs-core/blob/master/api/src/main/java/org/openmrs/api/impl/PatientServiceImpl.java#L1091), but the [exitFromCare](https://github.com/openmrs/openmrs-core/blob/52754eda69b824456ca0311f0cc7f2b83af5d59b/api/src/main/java/org/openmrs/api/impl/PatientServiceImpl.java#L972) implementation does not close any active programs.

cioan: @mogoodrich , I think the confusion is that there are 2 distinct exitFromCare implementations:

1. PatientService.[exitFromCare](https://github.com/openmrs/openmrs-core/blob/52754eda69b824456ca0311f0cc7f2b83af5d59b/api/src/main/java/org/openmrs/api/impl/PatientServiceImpl.java#L972) that only adds an Obs with reasonForExit Concept
2. LegacyUIService.[exitFromCare](https://github.com/openmrs/openmrs-module-legacyui/blob/master/api/src/main/java/org/openmrs/module/legacyui/api/impl/LegacyUIImpl.java#L67) that closes programs and discontinue drug orders

In this PR, I just used the LegacyUI exitFromCare method which I think it was the old way to exit a patient from care.

cioan: Thanks @mogoodrich for clarifying this over the phone. You are right, if for any other reason than Death we close all programs, then we should also close all active programs when the patient is marked as deceased. I have updated the PR and I will update the ticket too. Thanks!

mogoodrich (approved):
LGTM!

cioan:
Thanks @mogoodrich !

PR #177, opened 2022-01-28
LUI-184: Updates for log4j2.x

sivareddytw:
Ticket Id

[LUI-184](https://issues.openmrs.org/browse/LUI-184)

Talk Thread:
https://talk.openmrs.org/t/urgent-security-advisory-2021-12-11-re-apache-log4j-2/35341/16?u=grace

Description

We currently use log4j 1.x which is not supported anymore. It is EOL as described here https://blogs.apache.org/foundation/entry/apache_logging_services_project_announces

We can either update to its successor log4j 2

https://logging.apache.org/log4j/2.x/

### DIFF
--- a/omod/pom.xml
+++ b/omod/pom.xml
@@ -117,11 +117,6 @@
             <artifactId>jsp-api</artifactId>
         </dependency>
 
-		<dependency>
-			<groupId>org.apache.logging.log4j</groupId>
-			<artifactId>log4j-1.2-api</artifactId>
-		</dependency>
-
 	</dependencies>
 
 	<build>
--- a/omod/pom.xml
+++ b/omod/pom.xml
@@ -117,6 +117,11 @@
             <artifactId>jsp-api</artifactId>
         </dependency>
 
+		<dependency>
+			<groupId>org.apache.logging.log4j</groupId>
+			<artifactId>log4j-1.2-api</artifactId>
+		</dependency>
+
 	</dependencies>
 
 	<build>
--- a/omod/src/main/java/org/openmrs/web/controller/maintenance/ServerLogController.java
+++ b/omod/src/main/java/org/openmrs/web/controller/maintenance/ServerLogController.java
@@ -64,7 +64,7 @@ public class ServerLogController extends SimpleFormController {
 	 * @see org.springframework.web.servlet.mvc.AbstractFormController#formBackingObject(javax.servlet.http.HttpServletRequest)
 	 */
 	protected List<String> formBackingObject(HttpServletRequest request) throws ServletException {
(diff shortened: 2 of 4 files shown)


ibacher:
@sivareddytw Thanks for this. I can't merge it in as-is, because it will break the legacy UI on too many versions. That said, I think the fix needed here is much simpler. Just change [the condition in this line](https://github.com/openmrs/openmrs-module-legacyui/blob/cf00ec9c92f815ed5976c1541d8286014d6e99dc/omod/src/main/java/org/openmrs/web/controller/maintenance/ServerLogController.java#L67) to:

```java
ModuleUtil.matchRequiredVersions(OpenmrsConstants.OPENMRS_VERSION_SHORT, "2.1.5 - 2.1.*, 2.4.* - 2.*")
```

Note that it should still be safe to use the old `org.apache.log4j.Logger` classes because these will be replaced with the versions for the log4j2 compatibility layer.

sivareddytw:
@ibacher Thanks for review on this PR.

Just changing the condition is not removing dependency on log4j1.x. We still have [imports](https://github.com/openmrs/openmrs-module-legacyui/blob/cf00ec9c92f815ed5976c1541d8286014d6e99dc/omod/src/main/java/org/openmrs/web/controller/maintenance/ServerLogController.java#L25) from log4j1.x classes (eg: `org.apache.log4j.Logger`).

Upcoming OpenMRS 2.1.15 Release is not containing any log1.x dependency classes (eg: `org.apache.log4j.Logger`). 

I tried with OpenMRS 2.1.15-SNAPSHOT and LegacyUI with the change suggested and getting **java.lang.NoClassDefFoundError: org/apache/log4j/Logger** which is expected because `org.apache.log4j.Logger` from log4j1.x is not in the classpath. 

Idea is to remove dependency on log4j1.x completely and use loggers from log4j2.x.

ibacher:
@sivareddytw Unfortunately, we can't do that as we need to continue supporting versions of OpenMRS that use Log4J at least for the next few years. The solution here is to use Log4J2's Log4J 1.x compatibility library, which continues to be supported. I must've missed that this was excluded from the main PR. The artefact is the one [here](https://mvnrepository.com/artifact/org.apache.logging.log4j/log4j-1.2-api/2.17.1). Alternatively, I suppose, you could use [SLF4J's bridge for Log4J 1.x](https://mvnrepository.com/artifact/org.slf4j/log4j-over-slf4j/1.7.35).

To be clear, we *absolutely* cannot move these classes to using only Log4J2 classes as it will break all versions of OpenMRS prior to 2.4.x except 2.1.5. However, by replacing these classes with either of the bridge versions, Log4J 1.x is no longer used at all.

sivareddytw:
thanks @ibacher for detailed explanation. 
Updated PR with the feedback.
Let us know your comments.

ibacher (approved):

ibacher:
Thanks @sivareddytw!

PR #178, opened 2022-02-09
Revert "LUI-183: Fixed the concept initial value to ensure only the c…

cioan:
This is just reverting the commit that broke the causeOfDeath widget


cioan:
Thanks @dkayiwa !

PR #180, opened 2022-03-22
LUI-188 - Separate global properties for max encounter between encoun…

mseaton:
…ter and form entry tabs

### DIFF
--- a/omod/src/main/webapp/portlets/personFormEntry.jsp
+++ b/omod/src/main/webapp/portlets/personFormEntry.jsp
@@ -20,9 +20,14 @@ Parameters:
 
 <c:if test="${model.showDefinedNumberOfEncounters}">  <%-- showDefinedNumberOfEncounters shows the given number of encounters entered in the dashboard.encounters.maximumNumberToShow global property value --%>
 	<openmrs:hasPrivilege privilege="View Encounters">
-		<openmrs:globalProperty var="maxEncs" key="dashboard.encounters.maximumNumberToShow" defaultValue="" />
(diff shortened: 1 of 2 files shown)


omod/src/main/webapp/portlets/personFormEntry.jsp
```diff
@@ -20,9 +20,14 @@ Parameters:
...
 <c:if test="${model.showDefinedNumberOfEncounters}">  <%-- showDefinedNumberOfEncounters shows the given number of encounters entered in the dashboard.encounters.maximumNumberToShow global property value --%>
 	<openmrs:hasPrivilege privilege="View Encounters">
-		<openmrs:globalProperty var="maxEncs" key="dashboard.encounters.maximumNumberToShow" defaultValue="" />
-		<openmrs:portlet url="patientEncounters" id="patientDashboardEncounters" patientId="${patient.patientId}" parameters="num=${maxEncs}|title=Encounter.last.definedNumberOfEncounters|formEntryReturnUrl=${pageContext.request.contextPath}/patientDashboard.form" />
-		<br/>
+		<openmrs:globalProperty var="maxEncs" key="dashboard.formEntry.maximumNumberEncountersToShow" defaultValue="" />
+		<c:if test="${empty maxEncs}">
+			<openmrs:globalProperty var="maxEncs" key="dashboard.encounters.maximumNumberToShow" defaultValue="" />
+		</c:if>
+		<c:if test="${maxEncs != '0'}">
```
dkayiwa, line 27: @mseaton what has been the existing behaviour when `dashboard.encounters.maximumNumberToShow` was equal to `0`?

mseaton: It's a good question @dkayiwa , but my assumption is that no one would ever configure it this way, since it would also affect the encounters tab in a way that wouldn't make a lot of sense.  So my assumption is that this is only something someone would make 0 through the formEntry max encounters configuration.

PR #184, opened 2022-09-13
L UI 175-i identified where these global properties are used in by the legacyui module, and updated this module to gracefully handle missing values for these global properties, though some of the removed GPs were not having usecase in the implementation

jona42-ui:
**Description of what I changed**
Identified  where these global properties are used and update this module to gracefully handle missing values for these global properties. and i mainly found them in patientDashboardForm.jsp and patirntOverview.jsp

**this i PR  is based on issue;**
https://issues.openmrs.org/browse/LUI-175

### DIFF
--- a/omod/src/main/webapp/patientDashboardForm.jsp
+++ b/omod/src/main/webapp/patientDashboardForm.jsp
@@ -83,10 +83,18 @@
 		return false;
     }
 	
-	//hide the feature, cause of death if the global property is not defined.
+	//hide the feature(concepts that are not defined either by property or value).
+	window.onload = function showOrHide() {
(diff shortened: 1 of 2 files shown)


sherrif10:
Have you gone through different tickets and see how we can make a good ticket header.

jona42-ui:
> Have you gone through different tickets and see how we can make a good ticket header.

ooops! that was so long a header, have edited it. thanks @sherrif10

omod/src/main/webapp/patientDashboardForm.jsp
```diff
@@ -82,6 +82,19 @@
 		}
 		return false;
     }
+	
+	//hide the feature(concepts that are not defined either by property or value).
+	window.onload = function showOrHide() {
```
sherrif10, line 87: Well done on this. Do you mind sharing a small vedio or screenshots for this fix.

PR #185, opened 2022-12-02
LUI-192 - Provide ability to override user form and current users list

mseaton:


omod/src/main/java/org/openmrs/web/controller/maintenance/CurrentUsersController.java
```diff
@@ -30,26 +29,18 @@
...
 		log.debug("List current users");
 		if (!Context.hasPrivilege(PrivilegeConstants.GET_USERS)) {
 			throw new APIAuthenticationException("Privilege required: " + PrivilegeConstants.GET_USERS);
 		}
-		
-		modelMap.put("currentUsers", CurrentUsers.getCurrentUsernames(request.getSession()));
-		return "/module/legacyui/admin/maintenance/currentUsers";
+		return CurrentUsers.getCurrentUsernames(request.getSession());
 	}
-
```
mseaton, line 54: Because this existing controller is doing so little, I reverted it to an xml-based controller, which allows me to override this URL in my module by doing so in an annotated controller.  This page doesn't actually work anymore, but that's beside the point.

omod/src/main/java/org/openmrs/web/controller/user/UserController.java
```diff
@@ -0,0 +1,26 @@
...
+
+public class UserController extends SimpleFormController {
+	
+	/**
+	 * @see org.springframework.web.servlet.mvc.AbstractFormController#formBackingObject(HttpServletRequest)
+	 */
+	protected Object formBackingObject(HttpServletRequest request) throws ServletException {
+		return new HashMap<>();
+	}
+}
```
mseaton, line 26: This is just a dummy controller to back a new "user.htm" page which simply redirects to the user.form page.  By making this an xml-managed controller, and by changing just the link in the users.list page to point to this user.htm page instead of user.form, it maintains the existing behavior of opening user.form from users.list by default, but also allowing user.htm to be overridden in another module-defined annotated controller, thus opening a different user management page from the user.list results.

omod/src/main/java/org/openmrs/web/controller/user/UserFormController.java
```diff
@@ -72,12 +72,15 @@ public void initBinder(WebDataBinder binder) {
...
-		}
-		catch (Exception ex) {
-			log.error("Error while getting user", ex);
+		if (!StringUtils.isEmpty(userId)) {
+			try {
+				u = Context.getUserService().getUser(Integer.valueOf(userId));
+			}
+			catch (Exception ex) {
+				log.error("Error while getting user", ex);
+			}
```
mseaton, line 81: This has nothing to do with this feature, I just noticed a stack trace in the logs when I clicked on the link to create a new user (opens this form with userId = null), so put a null check around it.

omod/src/main/java/org/openmrs/web/dwr/OpenmrsDWRServlet.java
```diff
@@ -49,7 +49,7 @@ public void init(ServletConfig config) throws ServletException {
 	@Override
 	public void service(ServletRequest req, ServletResponse res) throws ServletException, IOException {
 		Thread.currentThread().setContextClassLoader(OpenmrsClassLoader.getInstance());
-
+
```
mseaton, line 52: Formatting change after build

mogoodrich (approved):
LGTM, defer to @dkayiwa  on some of the details.

PR #186, opened 2023-01-10
L UI-193, display program attribute value

cioan:
This is just a small change to display program attribute value if there are any associated with a program. It would look like this:

Screenshot 2023-01-10 at 4 30 01 PM

### DIFF
--- a/api/pom.xml
+++ b/api/pom.xml
@@ -15,7 +15,7 @@
 	<parent>
 		<groupId>org.openmrs.module</groupId>
 		<artifactId>legacyui</artifactId>
-		<version>1.13.0-SNAPSHOT</version>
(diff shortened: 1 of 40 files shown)


mogoodrich (approved):
Generally looks good to me, though I'd defer to @dkayiwa  as I haven't worked with this stuff in a long time.

The one thing I'd suggest would be to make sure c:forEach "items" is null-safe if you haven't already

cioan:
Thanks @mogoodrich ! Yes, c:forEach items is null-safe.
@dkayiwa , please let us know if this looks OK to you. Thanks!

cioan:
Thank you @dkayiwa !

PR #188, opened 2024-01-23
O3-2761: Add ability to save user email

jnsereko:
This has been done using java reflection API

Issue: https://openmrs.atlassian.net/browse/O3-2761

cc @dkayiwa @ibacher

### DIFF
--- a/omod/src/main/java/org/openmrs/web/controller/user/UserFormController.java
+++ b/omod/src/main/java/org/openmrs/web/controller/user/UserFormController.java
@@ -11,6 +11,7 @@ package org.openmrs.web.controller.user;
 
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
+import org.apache.commons.validator.EmailValidator;
 import org.openmrs.Person;
 import org.openmrs.PersonName;
 import org.openmrs.Provider;
@@ -43,6 +44,8 @@ import org.springframework.web.context.request.WebRequest;
 
 import javax.servlet.http.HttpServletResponse;
 import javax.servlet.http.HttpSession;
+
+import java.lang.reflect.Field;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Set;
@@ -112,6 +115,13 @@ public class UserFormController {
 	public String showForm(@RequestParam(required = false, value = "userId") Integer userId,
 	        @RequestParam(required = false, value = "createNewPerson") String createNewPerson,
 	        @ModelAttribute("user") User user, ModelMap model) {
+
+		try {
+			Field emailField = getEmailField("email", user);
+			model.addAttribute("email", emailField.get(user));
+		} catch (IllegalArgumentException | IllegalAccessException | NullPointerException e) {
+			log.warn("Email field not available for setting", e);
+		}
 		
 		// the formBackingObject method above sets up user, depending on userId and personId parameters
 		
@@ -157,6 +167,7 @@ public class UserFormController {
 	        @RequestParam(required = false, value = "roleStrings") String[] roles,
 	        @RequestParam(required = false, value = "createNewPerson") String createNewPerson,
 	        @RequestParam(required = false, value = "providerCheckBox") String addToProviderTableOption,
+	        @RequestParam(required = false, value = "email") String email,
 	        @ModelAttribute("user") User user, BindingResult errors, HttpServletResponse response) {
 		
 		UserService us = Context.getUserService();
@@ -227,6 +238,19 @@ public class UserFormController {
 				}
 			}
 			
+			//check validity of email then save it
+			if (isEmailValid(email)) {
+				Field emailField;
+				try {
+					emailField = getEmailField("email", user);
+					emailField.set(user, email);
+				} catch (IllegalArgumentException | IllegalAccessException | NullPointerException e) {
+					log.error("Error while setting the email field", e);
+				}
+			} else {
+				log.warn("Invalid or unspecified email: " + (email != null ? "'" + email + "'" : "not provided"));
+			}
+
 			Set<Role> newRoles = new HashSet<Role>();
 			if (roles != null) {
 				for (String r : roles) {
@@ -321,4 +345,27 @@ public class UserFormController {
 		return user == null ? true : user.getUserId() == null;
 	}
 	
+	/**
+	 * @return true if email is valid or false otherwise
+	 * @param email
+	 */
+	private boolean isEmailValid(String email) {
+		return StringUtils.hasLength(email) && EmailValidator.getInstance().isValid(email);
+	}
+	
+	/**
+	 * @return an email field
+	 * @param emailFieldName
+	 * @param user
+	 */
+	private Field getEmailField(String emailFieldName, User user) {
+		try {
+			Field emailField = user.getClass().getDeclaredField(emailFieldName);
+			emailField.setAccessible(true);
+			return emailField;
+		} catch (NoSuchFieldException | SecurityException e) {
+			log.warn("Email field not available for setting", e);
+			return null;
+		}
+	}
 }
--- a/omod/src/main/webapp/admin/users/userForm.jsp
+++ b/omod/src/main/webapp/admin/users/userForm.jsp
@@ -91,6 +91,10 @@
 				<openmrs:portlet url="nameLayout" id="namePortlet" size="full" parameters="layoutMode=edit|layoutShowTable=false|layoutShowExtended=false" />
 			</spring:nestedPath>
 			<tr>
+				<td><openmrs:message code="Email" /><span class="required">*</span></td>
+				<td><input type="email" size="30" name="email" value="${email}"/></td>
+			</tr>
+			<tr>
 				<td><openmrs:message code="Person.gender"/><span class="required">*</span></td>
 				<td><spring:bind path="user.person.gender">
 						<openmrs:forEachRecord name="gender">
--- a/omod/src/test/java/org/openmrs/web/controller/user/UserFormControllerTest.java
+++ b/omod/src/test/java/org/openmrs/web/controller/user/UserFormControllerTest.java
@@ -49,7 +49,7 @@ public class UserFormControllerTest extends BaseModuleWebContextSensitiveTest {
 		user.addName(new PersonName("This", "is", "Test"));
 		user.getPerson().setGender("F");
 		controller.handleSubmission(request, new MockHttpSession(), new ModelMap(), "", "Save User", "pass123", "pass123",
-		    null, null, null, new String[0], "true", null, user, new BindException(user, "user"),
+		    null, null, null, new String[0], "true", null, "", user, new BindException(user, "user"),
 		    new MockHttpServletResponse());
 	}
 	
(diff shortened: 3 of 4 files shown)


omod/src/test/java/org/openmrs/web/controller/user/UserFormControllerTest.java
```diff
@@ -49,7 +49,7 @@ public void handleSubmission_shouldWorkForAnExample() throws Exception {
 		user.addName(new PersonName("This", "is", "Test"));
 		user.getPerson().setGender("F");
 		controller.handleSubmission(request, new MockHttpSession(), new ModelMap(), "", "Save User", "pass123", "pass123",
-		    null, null, null, new String[0], "true", null, user, new BindException(user, "user"),
+		    null, null, null, new String[0], "true", null, "", user, new BindException(user, "user"),
```
dkayiwa, line 52: Can you explain the motivation behind adding that extra empty string parameter?

omod/src/main/webapp/admin/users/userForm.jsp
```diff
@@ -90,6 +90,10 @@
 			<spring:nestedPath path="user.person.names[0]">
 				<openmrs:portlet url="nameLayout" id="namePortlet" size="full" parameters="layoutMode=edit|layoutShowTable=false|layoutShowExtended=false" />
 			</spring:nestedPath>
+			<tr>
```
dkayiwa, line 93: Can you avoid displaying this for platform versions which do not have the email field?

omod/src/main/java/org/openmrs/web/controller/user/UserFormController.java
```diff
@@ -321,4 +345,27 @@ private Boolean isNewUser(User user) {
...
+	private boolean isEmailValid(String email) {
+		return StringUtils.hasLength(email) && EmailValidator.getInstance().isValid(email);
+	}
+	
+	/**
+	 * @return an email field
+	 * @param emailFieldName
+	 * @param user
+	 */
+	private Field getEmailField(String emailFieldName, User user) {
```
dkayiwa, line 361: Do you need to pass these as parameters? Do you plan to reuse this method for other parameters?

dkayiwa: Why not do something like this: https://github.com/openmrs/openmrs-sdk/blob/5.5.0/sdk-commons/src/main/java/org/openmrs/maven/plugins/utility/DefaultWizard.java#L805-L807
And then set the email field.

dkayiwa: Here is a second example: https://github.com/openmrs/openmrs-module-xforms/blob/4.3.14/omod/src/main/java/org/openmrs/module/xforms/page/controller/PatientRegPageController.java#L102

https://github.com/openmrs/openmrs-module-xforms/blob/4.3.14/omod/src/main/webapp/patientReg.jsp#L123

omod/src/main/webapp/admin/users/userForm.jsp
dkayiwa: Here is an example: https://github.com/openmrs/openmrs-module-xforms/blob/4.3.14/omod/src/main/java/org/openmrs/module/xforms/page/controller/PatientRegPageController.java#L102

https://github.com/openmrs/openmrs-module-xforms/blob/4.3.14/omod/src/main/webapp/patientReg.jsp#L123

omod/src/test/java/org/openmrs/web/controller/user/UserFormControllerTest.java
jnsereko: i have added a demo email

omod/src/main/webapp/admin/users/userForm.jsp
jnsereko: would this work 
```
try {
	Field emailField = getEmailField();
	model.addAttribute("email", emailField.get(user));
	model.addAttribute("hasEmailField", true);
} catch (IllegalArgumentException | IllegalAccessException | NullPointerException e) {
	log.warn("Email field not available for setting", e);
	model.addAttribute("hasEmailField", false);
}

```

omod/src/main/java/org/openmrs/web/controller/user/UserFormController.java
jnsereko: With this approach, i have used ` User.class.getDeclared...` I hope it is not a bad practice

omod/src/main/webapp/admin/users/userForm.jsp
```diff
@@ -90,6 +90,12 @@
 			<spring:nestedPath path="user.person.names[0]">
 				<openmrs:portlet url="nameLayout" id="namePortlet" size="full" parameters="layoutMode=edit|layoutShowTable=false|layoutShowExtended=false" />
 			</spring:nestedPath>
+			<c:if test="${hasEmailField}">
+				<tr>
+					<td><openmrs:message code="Email" /><span class="required">*</span></td>
```
dkayiwa, line 95: Is the email field required?

omod/src/main/java/org/openmrs/web/controller/user/UserFormController.java
```diff
@@ -112,6 +115,15 @@ public List<Role> getRoles(WebRequest request) {
 	public String showForm(@RequestParam(required = false, value = "userId") Integer userId,
 	        @RequestParam(required = false, value = "createNewPerson") String createNewPerson,
 	        @ModelAttribute("user") User user, ModelMap model) {
+
+		try {
+			Field emailField = getEmailField();
```
dkayiwa, line 120: Why don't you get the email field only for the platform versions that support it?

omod/src/main/webapp/admin/users/userForm.jsp
jnsereko: It is literally not required.  However, since reseting password actually depends on it, i thought it would be a good practice if we emphasise users to always fill it

dkayiwa: Required means we cannot save the user without it.

ibacher: > reseting password

I'm not sure I necessarily follow this. We don't (in the general case) have support for sending user's email for password resets, though it's possible for a module to add this capability.

omod/src/main/java/org/openmrs/web/controller/user/UserFormController.java
jnsereko: i was trying to avoid duplicating the code ` User.class.getDeclaredField("email")`  when; 1) determining if the platform version actually supports this field, and, 2) when i actually need access the email filed to set it

I have another way though.  It might not be ideal though  using `OPENMRS_VERSION_SHORT`
```
	private boolean isPlatformTwoPointTwoOrAbove() {
        String platformVersion = OpenmrsConstants.OPENMRS_VERSION_SHORT.substring(0, 3);

        try {
            float versionFloat = Float.parseFloat(platformVersion);
			if (versionFloat>=2.2) {
				return true;
			}
        } catch (NumberFormatException e) {
			log.error("Unable to parse platform value text to float", e);

        }
		return false;
    }
    ```

File: omod/src/main/webapp/admin/users/userForm.jsp (hunk 4 above)
Comment (jnsereko) [reply omod/src/main/webapp/admin/users/userForm.jsp]: @ibacher I am simply trying to make sure that [rest/v1/passwordreset](https://github.com/openmrs/openmrs-module-webservices.rest/blob/417b2775e1f4c0a11cf564bb3d7c5778337181b7/omod-2.2/src/main/java/org/openmrs/module/webservices/rest/web/v1_0/controller/openmrs2_2/PasswordResetController2_2.java#L32) can get working so that we can use it in the frontend  
This PR is might be followed by many more until this works

File: omod/src/main/java/org/openmrs/web/controller/user/UserFormController.java [hunk 6]
```diff
@@ -112,6 +115,17 @@ public List<Role> getRoles(WebRequest request) {
 	public String showForm(@RequestParam(required = false, value = "userId") Integer userId,
 	        @RequestParam(required = false, value = "createNewPerson") String createNewPerson,
 	        @ModelAttribute("user") User user, ModelMap model) {
+
+		try {
+			boolean isPlatformTwoPointTwoOrAbove = isPlatform22OrNewer();
+			model.addAttribute("hasEmailField", isPlatformTwoPointTwoOrAbove);
+			if (isPlatformTwoPointTwoOrAbove) {
```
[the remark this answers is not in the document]
jnsereko: I thought doing so would make us get `null` values for the `hasEmailField` most especially for platform versions below `2.2`
Honestly, i don't know how i can run platform below 2.2 with legacy so i am blindly coding here

omod/src/main/java/org/openmrs/web/controller/user/UserFormController.java
```diff
@@ -112,6 +115,17 @@ public List<Role> getRoles(WebRequest request) {
 	public String showForm(@RequestParam(required = false, value = "userId") Integer userId,
 	        @RequestParam(required = false, value = "createNewPerson") String createNewPerson,
 	        @ModelAttribute("user") User user, ModelMap model) {
+
+		try {
+			boolean isPlatform22OrNewer = isPlatform22OrNewer();
```
dkayiwa, line 120: Do we need the above variable? Why not just directly call the method in the if statement?

omod/src/main/java/org/openmrs/web/controller/user/UserFormController.java
```diff
@@ -227,6 +242,19 @@ public String handleSubmission(WebRequest request, HttpSession httpSession, Mode
 				}
 			}
 			
+			//check validity of email then save it
+			if (isPlatform22OrNewer() && isEmailValid(email)) {
+				Field emailField;
+				try {
+					emailField = getEmailField();
+					emailField.set(user, email);
+				} catch (IllegalArgumentException | IllegalAccessException | NullPointerException e) {
```
dkayiwa, line 251: Aren't these exceptions thrown and caught within getEmailField?

omod/src/main/java/org/openmrs/web/controller/user/UserFormController.java
```diff
@@ -227,6 +242,19 @@ public String handleSubmission(WebRequest request, HttpSession httpSession, Mode
 				}
 			}
 			
+			//check validity of email then save it
```
dkayiwa, line 245: I do not see any value added by the above comment.

omod/src/main/java/org/openmrs/web/controller/user/UserFormController.java
```diff
@@ -321,4 +349,43 @@ private Boolean isNewUser(User user) {
...
+	 */
+	private boolean isEmailValid(String email) {
+		return StringUtils.hasLength(email) && EmailValidator.getInstance().isValid(email);
+	}
+	
+	/**
+	 * @return an email field
+	 * @param user
+	 */
+	private Field getEmailField() {
```
dkayiwa, line 364: Shouldn't this method return the email field value instead of the field object?

jnsereko: i would find it hard to reuse this method it in `handleSubmission` if it was returning the value

dkayiwa: Oh i see!

jnsereko:
hey @dkayiwa  @ibacher anything else left here?

dkayiwa:
> hey @dkayiwa @ibacher anything else left here?

Did you respond to my review comments?

PR #190, opened 2024-03-30
TRUNK-6203: Global properties access should be privileged

wikumChamith:
After TRUNK-6203, users must possess the 'Get Global Properties' privilege to access global properties. This addition and removal of the global property are necessary for accessing global properties before a user logs in.

Ticket: https://openmrs.atlassian.net/browse/TRUNK-6203

### DIFF
--- a/omod/src/main/java/org/openmrs/web/servlet/LoginServlet.java
+++ b/omod/src/main/java/org/openmrs/web/servlet/LoginServlet.java
@@ -33,6 +33,7 @@ import org.openmrs.api.context.Context;
 import org.openmrs.api.context.ContextAuthenticationException;
 import org.openmrs.util.LocaleUtility;
 import org.openmrs.util.OpenmrsConstants;
+import org.openmrs.util.PrivilegeConstants;
 import org.openmrs.web.OpenmrsCookieLocaleResolver;
 import org.openmrs.web.WebConstants;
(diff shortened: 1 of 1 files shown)


dkayiwa:
> Strangely, it sometimes works without these changes 😅

Not a good sign. 😊

wikumChamith:
> > Strangely, it sometimes works without these changes 😅
> 
> Not a good sign. 😊

That happened only one time probably a issue on my side.

omod/src/main/java/org/openmrs/web/servlet/LoginServlet.java
```diff
@@ -81,8 +82,16 @@ protected void doPost(HttpServletRequest request, HttpServletResponse response)
 		// look up the allowed # of attempts per IP
 		Integer allowedLockoutAttempts = 100;
 		
-		String allowedLockoutAttemptsGP = Context.getAdministrationService().getGlobalProperty(
-		    GP_ALLOWED_LOGIN_ATTEMPTS_PER_IP, "100");
+		String allowedLockoutAttemptsGP = "";
```
dkayiwa, line 85: Why not initialise this with the default value?

wikumChamith: I updated the PR.

PR #196, opened 2024-08-27
LUI-197 UI for saving concept reference ranges

dicksonmulli:
https://openmrs.atlassian.net/browse/LUI-197


dkayiwa:
Do you have a screenshot of how this looks like?

dicksonmulli:
`Screenshot` : When no reference ranges added:

Screenshot 2024-09-02 at 16 43 53


`Screenshot`: Reference ranges:

Screenshot 2024-09-02 at 16 45 40

`Screenshot`: Concept screen:

Screenshot 2024-09-02 at 16 46 45

omod/src/main/java/org/openmrs/web/controller/ConceptFormController.java
```diff
@@ -662,6 +668,11 @@ public Concept getConceptFromFormData() {
 				cn.setDisplayPrecision(displayPrecision);
 				cn.setUnits(units);
 				
+				for (ConceptReferenceRange referenceRange : this.referenceRanges) {
+					referenceRange.setConceptNumeric(cn);
```
dkayiwa, line 672: What leads to the state where a concept numeric is loaded without the reference ranges?

dicksonmulli: An error will occur because of lack of foreign key.

dkayiwa: If you try run that code in debug mode, you will notice that the reference range already has the concept numeric property set.

dkayiwa:
@dicksonmulli in the hibernate mapping file, did you try map concept reference ranges the same way as concept mappings?

dicksonmulli:
> @dicksonmulli in the hibernate mapping file, did you try map concept reference ranges the same way as concept mappings?

Yes, but I am still getting the same error : `org.hibernate.exception.ConstraintViolationException: could not execute batch`

dkayiwa:
> Yes, but I am still getting the same error : `org.hibernate.exception.ConstraintViolationException: could not execute batch`

What value do you have in the `cascade` attribute for `referenceRanges`?

dicksonmulli:
> What value do you have in the `cascade` attribute for `referenceRanges`?

I changed to `all,delete-orphan,evict` but I was experiencing the same error. I however got it to work by adding `inverse="true"` 👍

omod/src/main/java/org/openmrs/web/controller/ConceptFormValidator.java
```diff
@@ -113,4 +119,78 @@ else if (backingObject.getNamesByLocale().get(locale).getConceptNameId() != null
...
+			int index = 0;
+			for (ConceptReferenceRange referenceRange : referenceRanges) {
+				
+				if (referenceRange.getId() == null) {
+					if (referenceRange.getHiAbsolute() == null) {
+						setReferenceRangeErrors(errors, index, "hiAbsolute",
+						    "Concept.referenceRanges.error.high.absolute.value.required",
+						    "Concept.referenceRanges.error.absolute.value.required");
+					} else {
+						if (referenceRange.getHiAbsolute() > conceptNumeric.getHiAbsolute()) {
```
dkayiwa, line 148: What happens when conceptNumeric.getHiAbsolute() is null?

dicksonmulli: Fixed

dkayiwa:
@dicksonmulli i think we have one more remaining bug here.
Add a first reference range row, then add a second one, and then remove the first one, finally save the concept. You will get an ugly error.

dicksonmulli:
> @dicksonmulli i think we have one more remaining bug here. Add a first reference range row, then add a second one, and then remove the first one, finally save the concept. You will get an ugly error.

Fixed now.

dkayiwa:
@dicksonmulli did you forget to make the commit?

dkayiwa:
@dicksonmulli i have noticed that the order of insertion of reference ranges is preserved only when they are added to an existing concept. But when you create a new one and immediately add the reference ranges before saving it, the order is not preserved.

dkayiwa:
@dicksonmulli i also noticed that when i run this on platform versions before 2.7, i get this in the logs when i save a concept: https://pastebin.com/wcDCXc8P

Secondly, i also see the `Reference Ranges` label together with the `Add a Reference Range` button yet these are supposed to show only on platform 2.7 and above.

dicksonmulli:
> i also noticed that when i run this on platform versions before 2.7, i get this in the logs when i save a concept: https://pastebin.com/wcDCXc8P

Yes, this is is the expected behaviour because we are catching the exception incase the version is not compatible - so that the application will not fail to run. However, we can change the log from `error` to `info` if that is okay.

dkayiwa:
Two things:

1) The add reference ranges label and button should not be visible to lower versions of the platfoem.

2) There is totally nothing wrong with someone running versions of the platform lower than 2.7 and hence it does not add value to clutter logs with this, even a mere logging of info does not help in any way.

omod/src/main/java/org/openmrs/web/controller/ConceptFormController.java
```diff
@@ -678,6 +697,161 @@ public Concept getConceptFromFormData() {
...
+		 * @param referenceRange referenceRange
+		 * @since 1.17.0
+		 */
+		private void removeReferenceRange(ConceptNumeric cn, ConceptReferenceRange referenceRange) {
+			try {
+				Object platformReferenceRange = new ConceptFormMapper().mapToConceptReferenceRange(referenceRange, cn);
+				setMethodValue(cn, "removeReferenceRange", platformReferenceRange);
+			}
+			catch (Exception exception) {
+				logger.warn("Failed to remove concept reference range. "
```
dkayiwa, line 742: The logs will be filled with many of these warnings for those running platform versions lower than 2.7
I do not find it a good practice to fill logs with such warning where there is completely nothing wrong about running lower versions of the platform.

omod/src/main/webapp/admin/dictionary/conceptForm.jsp
```diff
@@ -815,6 +820,65 @@
 			</table>
 		</td>
 	</tr>
+	<c:if test="${fn:substring(openmrsVersion, 0, 5) >= '2.7.0'}">
```
dkayiwa, line 823: Can we shield the client from this brittle logic by having something along these lines:
https://github.com/openmrs/openmrs-module-xforms/blob/master/api/src/main/java/org/openmrs/module/xforms/util/XformsUtil.java#L841

https://github.com/search?q=org%3Aopenmrs+isOnePointNineOneAndAbove&type=code

dkayiwa: Then you can just have a test in the jsp like this: https://github.com/search?q=repo%3Aopenmrs%2Fopenmrs-module-xforms%20useOpenmrsMessageTag&type=code

api/src/main/java/org/openmrs/module/legacyui/GeneralUtils.java
```diff
@@ -68,4 +69,49 @@ public static Concept getConcept(String id) {
...
 		return cpt;
 	}
+	
+	/**
+	 * Checks if current version of openmrs is greater or equal to 2.7.0
+	 * 
+	 * @return true if current version is greater or equal to 2.7.0 and false otherwise
+	 * @since 1.17.0
+	 */
+	public static boolean isTwoPointSevenAndAbove() {
```
dkayiwa, line 79: This is cimplicated and brittle. Why don't you use the same trick of trying to load a class like concept reference range which was added in 2.7?

dicksonmulli: This boolean is used in UI/JSP hence the need to check the version instead of loading a particular class. I think this will be reliable as long as we have the version constant(s).

dkayiwa: We can still have a boolean with loading a particular class.

api/src/main/java/org/openmrs/module/legacyui/GeneralUtils.java
```diff
@@ -68,4 +69,24 @@ public static Concept getConcept(String id) {
...
+	 * ConceptReferenceRange class is loaded, then the current version is greater than or equal to
+	 * 2.7.0
+	 * 
+	 * @return true if current version is greater or equal to 2.7.0 and false otherwise
+	 * @since 1.17.0
+	 */
+	public static boolean isTwoPointSevenAndAbove() {
+		try {
+			Class<?> referenceRangeClass = Class.forName("org.openmrs.ConceptReferenceRange");
+			referenceRangeClass.getMethod("getCriteria");
```
dkayiwa, line 85: What value does line 85 add?

dkayiwa:
@dicksonmulli i have noticed that with the changes in this pull request, when i am creating a new concept, the `Set Members` textarea is displayed even before clicking the `Is Set` checkbox. Do you mind correcting that?

dkayiwa:
@dicksonmulli and FWIW, this incorrect display happens only on platform versions lower than 2.7

dkayiwa:
@dicksonmulli now when a concept is saved with the `Is Set` checkbox clicked, on opening it again, the textarea is no longer displayed by default.

PR #197, opened 2024-09-07
resolve 3.4.1 by eliminating access to seeing errors without authentication

slubwama:
Paired with PR https://github.com/openmrs/openmrs-core/pull/4742 in core

### DIFF
--- a/api/src/main/resources/messages.properties
+++ b/api/src/main/resources/messages.properties
@@ -6,6 +6,7 @@ legacyui.manageuser.providerAccount=Provider Account
 legacyui.manageuser.createProviderAccount=Create a Provider account for this user
 legacyui.manageuser.providerIdentfier=Provider Identifier(s):
 legacyui.manageuser.noProviderIdentifier=No Identifier Specified
+legacyui.error.aunthenticationRequired=Authentication Required
 
 ${project.parent.artifactId}.Location.purgeLocation=Permanently Delete Location
 ${project.parent.artifactId}.Location.confirmDelete=Are you sure you want to delete this Location? It will be permanently removed from the system.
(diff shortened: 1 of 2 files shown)


dkayiwa:
Can you attach screenshots of how this looks like with and without authentication?

slubwama:
> Can you attach screenshots of how this looks like with and without authentication?

Screenshot 2024-09-09 at 03 53 25

omod/src/main/webapp/uncaughtException.jsp
```diff
@@ -51,195 +55,207 @@
...
 
 </div>
-	
+
+
+<%
+    } else {
+
+%>
+<h2><openmrs:message code="no.authenticated.user.found"/></h2>
```
dkayiwa, line 254: This error message looks like the action was about looking for an authenticated user. What do you think of using something like `Authentication Required`?

slubwama: @dkayiwa resolved

api/src/main/resources/messages.properties
```diff
@@ -6,6 +6,7 @@ legacyui.manageuser.providerAccount=Provider Account
 legacyui.manageuser.createProviderAccount=Create a Provider account for this user
 legacyui.manageuser.providerIdentfier=Provider Identifier(s):
 legacyui.manageuser.noProviderIdentifier=No Identifier Specified
+legacyui.error.aunthenticationRequired=Authentication Required
```
ibacher, line 9: We should probably fix the spelling here `authenticationRequired`

dkayiwa:
@slubwama i see something like `(The full error stack trace output is in the source of this page.)`
But i do not see it when i check the page source.

PR #198, opened 2024-09-24
Redirecting to the login page if authentication is needed.

wikumChamith:


omod/src/main/webapp/uncaughtException.jsp
```diff
@@ -253,7 +253,8 @@
 %>
 <h2><openmrs:message code="legacyui.error.aunthenticationRequired"/></h2>
 <%
-    }
+		response.sendRedirect("login.htm");
```
dkayiwa, line 256: I think it is this commit which needs to be improved: https://github.com/openmrs/openmrs-core/commit/e7eb6ffb566fdb2bbad2672110447cb6ac552fd0

wikumChamith: I tried reverting changes in this PR but that didn't work.

dkayiwa: Does that mean that this https://github.com/openmrs/openmrs-core/commit/e7eb6ffb566fdb2bbad2672110447cb6ac552fd0 is not the cause of that problem?

wikumChamith: If that's the case, wouldn't reverting the change resolve the issue?

dkayiwa: I have not tried it out. I am just asking to find out what you have discovered.

wikumChamith: It didn't work for me.

dkayiwa: Did you try to look into what commit triggered it?

dkayiwa: Can we improve that legacyui module commit such that it accomplishes its intended goal without introducing this problem?

wikumChamith: Doesn't this achieve its intended goal? This Redirect unauthenticated users to the login page rather than displaying an error.

dkayiwa: What happens when login is provided by a page different from login.htm?

wikumChamith: We use this in other places on the code too.

https://github.com/openmrs/openmrs-module-legacyui/blob/08f44dcee71ed61d14bddc15a7ac1474e15b4305/omod/src/main/webapp/findPatient.jsp#L3

https://github.com/openmrs/openmrs-module-legacyui/blob/08f44dcee71ed61d14bddc15a7ac1474e15b4305/omod/src/main/webapp/authorizationHandlerInclude.jsp#L35

dkayiwa: You have a very strong point. 😊

PR #201, opened 2024-09-30
Fix search data showing null

wikumChamith:


omod/src/main/webapp/resources/scripts/jquery-ui/js/openmrsSearch.js
```diff
@@ -839,7 +839,7 @@ function OpenmrsSearch(div, showIncludeVoided, searchHandler, selectionHandler,
             var cols = this.options.fieldsAndHeaders;
             rRowData = $j.map(cols, function(c) {
                 var data = rowData[c.fieldName];
-                if(data == null)
```
dkayiwa, line 842: Have you looked into why we did not have this problem originally without this change?

wikumChamith: Now we return null as a string. I am not sure where we made that change though.

dkayiwa: Can you try to locate where we made that change?

wikumChamith: This is what is causing the issue: https://github.com/openmrs/openmrs-module-legacyui/blob/f29296d63bcc985ee9dd80301e5fd164aef83c72/omod/src/main/java/org/openmrs/web/dwr/PersonListItem.java#L120

WebUtil.escapeHTML() converts null to a string.

This was introduced here: https://github.com/openmrs/openmrs-module-legacyui/pull/191/files

PR #208, opened 2024-11-05
Escape HTML inn Process of creating patient to avoid XSS

slubwama:


dkayiwa:
@slubwama were you able to reproduce the problem that was reported of doing this without going through the UI?

slubwama:
> @slubwama were you able to reproduce the problem that was reported of doing this without going through the UI?

@dkayiwa given time. Yes Got the tool still trying to learn how to use it to replicate the error. But from the look of things basing on the comments Its caused as a results of not escaping the attributes  items in patient object

dkayiwa:
@slubwama this time round, i do not feel comfortable merging pull requests, if we cannot first of all reproduce the problem and then afterwords confirm that with the changes in the pull request, the same exact steps can no longer reproduce the reported problem. Do i make sense?

PR #209, opened 2024-11-13
(Enhancement) Optimize patient dashboard loading by implementing pagination for observations

ODORA0:
Description:
* Implemented pagination for patient observations to improve dashboard loading performance
* Added pagination parameters (page size and current page) to the model
* Optimized weight and height observation retrieval by fetching them directly
* Added conditional check for visits feature before loading visit data
* Improved BMI calculation with better error handling
* Added utility methods for unit conversion and pagination parameter handling

Before this change, the dashboard would load all observations at once, causing performance issues for patients with many observations. This implementation:
- Loads observations in configurable page sizes (default: 50)
- Allows pagination through URL parameters (?page=0&pageSize=50)
- Maintains all existing functionality while improving performance
- Adds pagination metadata to support UI implementation
- Optimizes memory usage by loading observations in chunks

Testing:
- Verified dashboard loads faster with paginated observations
- Tested with patients having large numbers of observations
- Confirmed BMI calculation works correctly with different unit types
- Validated visit loading with feature flag enabled/disabled

To configure:
Add `dashboard.defaultPageSize` to global properties to set default page size (optional)

### DIFF
--- a/omod/src/main/java/org/openmrs/web/controller/PortletController.java
+++ b/omod/src/main/java/org/openmrs/web/controller/PortletController.java
@@ -11,12 +11,14 @@ package org.openmrs.web.controller;
 
 import java.io.IOException;
 import java.util.ArrayList;
+import java.util.Comparator;
 import java.util.Date;
 import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.UUID;
+import java.util.Collections;
 
 import javax.servlet.ServletException;
 import javax.servlet.http.HttpServletRequest;
@@ -39,6 +41,7 @@ import org.openmrs.api.AdministrationService;
 import org.openmrs.api.ConceptService;
 import org.openmrs.api.context.Context;
 import org.openmrs.module.legacyui.GeneralUtils;
+import org.openmrs.util.OpenmrsConstants;
 import org.openmrs.util.PrivilegeConstants;
 import org.openmrs.web.WebConstants;
 import org.springframework.util.StringUtils;
@@ -48,6 +51,8 @@ import org.springframework.web.servlet.mvc.Controller;
 public class PortletController implements Controller {
 	
 	protected Log log = LogFactory.getLog(this.getClass());
+	
+	private static final int DEFAULT_PAGE_SIZE = 50;
 	
 	/**
 	 * This method produces a model containing the following mappings:
@@ -97,11 +102,11 @@ public class PortletController implements Controller {
 	 */
 	@SuppressWarnings("unchecked")
 	public ModelAndView handleRequest(HttpServletRequest request, HttpServletResponse response) throws ServletException,
-	        IOException {
-		
+			IOException {
+
 		AdministrationService as = Context.getAdministrationService();
 		ConceptService cs = Context.getConceptService();
-		
+
 		// find the portlet that was identified in the openmrs:portlet taglib
 		Object uri = request.getAttribute("javax.servlet.include.servlet_path");
 		String portletPath = "";
@@ -112,8 +117,8 @@ public class PortletController implements Controller {
 			String lastRequestId = (String) session.getAttribute(WebConstants.OPENMRS_PORTLET_LAST_REQ_ID);
 			if (uniqueRequestId.equals(lastRequestId)) {
 				model = (Map<String, Object>) session.getAttribute(WebConstants.OPENMRS_PORTLET_CACHED_MODEL);
-				
-				// remove cached parameters 
+
+				// remove cached parameters
 				List<String> parameterKeys = (List<String>) model.get("parameterKeys");
 				if (parameterKeys != null) {
 					for (String key : parameterKeys) {
@@ -128,26 +133,27 @@ public class PortletController implements Controller {
 				session.setAttribute(WebConstants.OPENMRS_PORTLET_CACHED_MODEL, model);
 			}
 		}
-		
+
 		if (uri != null) {
 			long timeAtStart = System.currentTimeMillis();
 			portletPath = uri.toString();
-			
+
 			// Allowable extensions are '' (no extension) and '.portlet'
 			if (portletPath.endsWith("portlet")) {
 				portletPath = portletPath.replace(".portlet", "");
 			} else if (portletPath.endsWith("jsp")) {
 				throw new ServletException(
-				        "Illegal extension used for portlet: '.jsp'. Allowable extensions are '' (no extension) and '.portlet'");
+						"Illegal extension used for portlet: '.jsp'. Allowable extensions are '' (no extension) and '.portlet'");
 			}
-			
+
 			log.debug("Loading portlet: " + portletPath);
-			
+
 			String id = (String) request.getAttribute("org.openmrs.portlet.id");
 			String size = (String) request.getAttribute("org.openmrs.portlet.size");
 			Map<String, Object> params = (Map<String, Object>) request.getAttribute("org.openmrs.portlet.parameters");
-			Map<String, Object> moreParams = (Map<String, Object>) request.getAttribute("org.openmrs.portlet.parameterMap");
-			
+			Map<String, Object> moreParams = (Map<String, Object>) request
+					.getAttribute("org.openmrs.portlet.parameterMap");
+
 			model.put("now", new Date());
 			model.put("id", id);
 			model.put("size", size);
@@ -160,14 +166,15 @@ public class PortletController implements Controller {
 				parameterKeys.addAll(moreParams.keySet());
 			}
 			model.put("parameterKeys", parameterKeys); // so we can clean these up in the next request
-			
-			// if there's an authenticated user, put them, and their patient set, in the model
+
+			// if there's an authenticated user, put them, and their patient set, in the
+			// model
 			if (Context.getAuthenticatedUser() != null) {
 				model.put("authenticatedUser", Context.getAuthenticatedUser());
 			}
-			
+
 			Integer personId = null;
-			
+
 			// if a patient id is available, put patient data documented above in the model
 			Object o = request.getAttribute("org.openmrs.portlet.patientId");
 			if (o != null) {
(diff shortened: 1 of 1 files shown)


dkayiwa:
Did you get a chance to take a look at this? https://openmrs.atlassian.net/wiki/spaces/docs/pages/25477199/Pull+Request+Tips

dkayiwa:
Can we also include the ticket id in the commit message?

omod/src/main/java/org/openmrs/web/controller/PortletController.java
```diff
@@ -195,8 +198,24 @@ public ModelAndView handleRequest(HttpServletRequest request, HttpServletRespons
 					}
 					
 					if (Context.hasPrivilege(PrivilegeConstants.GET_OBS)) {
-						List<Obs> patientObs = Context.getObsService().getObservationsByPerson(p);
-						model.put("patientObs", patientObs);
+						// Get pagination parameters
+						Integer page = getPageParameter(request);
+						Integer pageSize = getPageSizeParameter(request, as);
+						
+						// Get all observations
```
dkayiwa, line 205: For each time that one moves through the pages, you are still fetching and loading all the patient's observations in memory. My impression was that you would load in memory only those observations for the current page.

omod/src/main/java/org/openmrs/web/controller/PortletController.java
```diff
@@ -195,8 +202,38 @@ public ModelAndView handleRequest(HttpServletRequest request, HttpServletRespons
...
+						List<PERSON_TYPE> personTypes = null;
+						List<Location> locations = null;
+						List<String> sort = Collections.singletonList("obsDatetime desc");
+						Integer obsGroupId = null;
+						Date fromDate = null;
+						Date toDate = null;
+						boolean includeVoided = false;
+						
+						// Get observations for the current page
+						List<Obs> paginatedObs = Context.getObsService().getObservations(persons, encounters, questions,
```
dkayiwa, line 225: Though the comment says get observations for the current page, i do not seen anything in the getObservations method call that does so. Was it just a typo?

ODORA0: Ooh yeah, my bad the getObservations() method doesn't have parameters for offset/limit pagination, let me remove that comment

dkayiwa: So then, shall we then have any dashboard loading optimisation if the method loads all of them in memory?

ODORA0: Okay, I critically looked at the ObsService interface, didn't notice that it lacks proper pagination parameters (offset/limit). Maybe this should be done as a DAO level implementation so we can have a DB level pagination support, but I would think of doing it this way as well i.e. we can use an existing API features like mostRecentN to limit initial load like so

```suggestion
List<Obs> paginatedObs = Context.getObsService().getObservations(
    persons,
    null,
    null,
    null,
    null,
    null,
    Collections.singletonList("obsDatetime desc"),
    pageSize,  // mostRecentN - only fetch what we need
    null,
    null,
    null,
    false
);
```

or using date filtering

```suggestion
Calendar cal = Calendar.getInstance();
cal.add(Calendar.DAY_OF_MONTH, -30);
Date thirtyDaysAgo = cal.getTime();

List<Obs> recentObs = Context.getObsService().getObservations(
    persons,
    null,
    null,
    null,
    null,
    null,
    sort,
    null,
    null,
    thirtyDaysAgo,
    new Date(),
    false
);
```

Your thoughts @dkayiwa

ODORA0: Okay!
@eudson Any thoughts on this, is it acceptable to only show the most recent observations?

omod/src/main/java/org/openmrs/web/controller/PortletController.java
```diff
@@ -190,7 +193,10 @@ public ModelAndView handleRequest(HttpServletRequest request, HttpServletRespons
 					
 					// add encounters if this user can view them
 					if (Context.hasPrivilege(PrivilegeConstants.GET_ENCOUNTERS)) {
-						model.put("patientEncounters", Context.getEncounterService().getEncountersByPatient(p));
+						model.put(
+						    "patientEncounters",
+						    Context.getEncounterService().getEncounters(p.getPatientIdentifier().getIdentifier(), 0, 100,
```
eudson, line 198: @ODORA0 why are you hardcoding the start(0) and length(100) ? My understanding of these two variables is that they can help paginate the results, the `start` being the page and the `length` the number of records per page.

omod/src/main/java/org/openmrs/web/controller/PortletController.java
```diff
@@ -202,36 +208,16 @@ public ModelAndView handleRequest(HttpServletRequest request, HttpServletRespons
...
-						
-						// Get observations for the current page
-						List<Obs> paginatedObs = Context.getObsService().getObservations(persons, encounters, questions,
-						    answers, personTypes, locations, sort, null, obsGroupId, fromDate, toDate, includeVoided);
-						
-						// Get total count for pagination
-						Integer totalCount = Context.getObsService().getObservationCount(persons, encounters, questions,
-						    answers, personTypes, locations, obsGroupId, fromDate, toDate, includeVoided);
+						List<Obs> paginatedObs = Context.getObsService().getObservations(persons, null, null, null, null,
+						    null, Collections.singletonList("obsDatetime desc"), pageSize,
```
eudson, line 217: Can we apply the same concept here?

eudson (changes requested):

omod/src/main/java/org/openmrs/web/controller/PortletController.java
ODORA0: So I am using the using proper ObsService method signature for observations and using mostRecentN for pagination with the pageSize parameter, using same pageSize from global property or default which is set to 50 thus showing most recent observations up to pageSize limit

omod/src/main/java/org/openmrs/web/controller/PortletController.java
ODORA0: My bad here, I might have forgotten to make the change but changed so that we using the proper pagination parameters using the configured page size in the `maximumNumberToShow` GP or default to only 50 when empty and allow users to paginate through encounters.

omod/src/main/java/org/openmrs/web/controller/PortletController.java
```diff
@@ -399,30 +479,56 @@ public ModelAndView handleRequest(HttpServletRequest request, HttpServletRespons
...
+	 * that context
 	 * could be null when this method is called.
 	 */
 	protected void populateModel(HttpServletRequest request, Map<String, Object> model) {
 	}
-	
+
+	private Integer getPageParameter(HttpServletRequest request) {
+		try {
+			return Integer.parseInt(request.getParameter("page"));
```
dkayiwa, line 511: Can you take a look at our naming convention for the pagination parameters? https://openmrs.atlassian.net/wiki/spaces/docs/pages/25469830/REST+Web+Services+API+For+Clients#Limiting-the-number-of-results-and-paging

omod/src/main/java/org/openmrs/web/controller/PortletController.java
```diff
@@ -399,30 +479,56 @@ public ModelAndView handleRequest(HttpServletRequest request, HttpServletRespons
...
+	}
+
+	private Integer getPageSizeParameter(HttpServletRequest request, AdministrationService as) {
+		try {
+			String pageSizeParam = request.getParameter("pageSize");
+			if (pageSizeParam != null) {
+				return Integer.parseInt(pageSizeParam);
+			}
+
+			String globalPageSize = as.getGlobalProperty("dashboard.encounters.maximumNumberToShow");
```
dkayiwa, line 524: Is this pull request about encounters or observations?

dkayiwa: And just in case you find this helpful: https://openmrs.atlassian.net/browse/LUI-188

omod/src/main/java/org/openmrs/web/controller/PortletController.java
dkayiwa: Basing on the user interface screen that this controller serves, could you remind me on how this parameter is changed to a different value?

ODORA0: So the method retrieves the page parameter from the HTTP request, and if it'ss missing or is invalid like non numeric it will default to 0. The page value calculates which subset of data (encounters or observations) to fetch `patientDashboard?page=2`

dkayiwa: What i mean is, how is this parameter changed in the HTTP request by the user interface?

omod/src/main/java/org/openmrs/web/controller/PortletController.java
ODORA0: Yes, we thought we could align encounters as well to limit it to based max numbers to show or default to the fallback property 50. cc @eudson  
Encounters fetched are now limited to the value of the GP

omod/src/main/java/org/openmrs/web/controller/PortletController.java
ODORA0: With the dropdown selector

dkayiwa: Does it select page or pageSize?

omod/src/main/java/org/openmrs/web/controller/PortletController.java
ODORA0: Okay so that solution effectively decouples the configuration for the Form Entry tab and the Encounters tab which tackles both usability and configurability needs without disrupting existing any setups. Are you trying to say we employ the same thing but for the patientEncounters jsp, LOL, I am having a brain fart right now with looking at old tech am not as familiar with

omod/src/main/java/org/openmrs/web/controller/PortletController.java
dkayiwa: Can you share a screenshot of that entire page?

ODORA0: image

omod/src/main/java/org/openmrs/web/controller/PortletController.java
dkayiwa: One should be able to set different values for encounters and observations.

omod/src/main/java/org/openmrs/web/controller/PortletController.java
dkayiwa: Awesome!!! 👍

dkayiwa: Did you check the link about naming conventions?

ODORA0: Yes, I have changed that, please review

omod/src/main/java/org/openmrs/web/controller/PortletController.java
ODORA0: This involves creating a different GP for Obs as in `maximumNumberObservationsToShow`? Please advise

dkayiwa: Yes you need to create a separate global property for it.

dkayiwa:
Can you just avoid running it?

ODORA0:
@dkayiwa Done

omod/src/main/resources/config.xml
```diff
@@ -382,6 +382,14 @@
 			Allows one to limit the number of encounters shown on the form entry tab of the patient dashboard specifically
 		</description>
 	</globalProperty>
+		<globalProperty>
+		<property>dashboard.observations.maximumNumberToShow</property>
+		<defaultValue>false</defaultValue>
```
dkayiwa, line 387: The above default value does not look valid.

ODORA0: Changed

omod/src/main/java/org/openmrs/web/controller/PortletController.java
```diff
@@ -195,43 +206,67 @@ public ModelAndView handleRequest(HttpServletRequest request, HttpServletRespons
...
 					if (Context.hasPrivilege(PrivilegeConstants.GET_OBS)) {
-						List<Obs> patientObs = Context.getObsService().getObservationsByPerson(p);
-						model.put("patientObs", patientObs);
+						Integer limit = getLimitParameter(request, as, "dashboard.observations.maximumNumberToShow");
+						
+						Person person = (Person) p;
+						List<Person> persons = Collections.singletonList(person);
+						
+						// Get most recent observations using limit parameter
+						List<Obs> paginatedObs = Context.getObsService().getObservations(persons, null, null, null, null,
```
dkayiwa, line 215: According to the user interface you shared, what do you do with the startIndex?

ODORA0: Select the page size

dkayiwa: Is the page size the same as startIndex?

ODORA0: Yes, based on the link you shared I renamed pageSize to that

dkayiwa: Where are you passing the two parameters in the getObservations api call?

ODORA0: We are only passing the limit and not startIndex directly in that method call

dkayiwa: Which means that if you are on the last page, you still load all obs from the database to memory.

ODORA0: @dkayiwa Yes that's right, added that

omod/src/main/java/org/openmrs/web/controller/PortletController.java
```diff
@@ -195,43 +206,68 @@ public ModelAndView handleRequest(HttpServletRequest request, HttpServletRespons
...
+						Integer startIndex = getStartIndexParameter(request);
+						
+						Person person = (Person) p;
+						List<Person> persons = Collections.singletonList(person);
+						
+						// Get most recent observations using limit parameter
+						List<Obs> paginatedObs = Context.getObsService().getObservations(persons, null, null, null, null, null, 
+						Collections.singletonList("obsDatetime desc"), limit, startIndex, null, null, false);					
+						
+						model.put("limit", limit);
```
dkayiwa, line 219: Can you remind me of why we are storing `limit` in the model?

omod/src/main/java/org/openmrs/web/controller/PortletController.java
```diff
@@ -183,7 +188,13 @@ public ModelAndView handleRequest(HttpServletRequest request, HttpServletRespons
 					
 					// add encounters if this user can view them
 					if (Context.hasPrivilege(PrivilegeConstants.GET_ENCOUNTERS)) {
-						model.put("patientEncounters", Context.getEncounterService().getEncountersByPatient(p));
+						Integer limit = getLimitParameter(request, as, "dashboard.encounters.maximumNumberToShow");
+						Integer startIndex = getStartIndexParameter(request);
+						
+						model.put(
+						    "patientEncounters",
+						    Context.getEncounterService().getEncounters(p.getPatientIdentifier().getIdentifier(),
```
dkayiwa, line 196: We should be using `Context.getEncounterService()getEncounters(EncounterSearchCriteria encounterSearchCriteria)`

dkayiwa, line 196: Why not use the more efficient? https://github.com/openmrs/openmrs-core/blob/2.0.0/api/src/main/java/org/openmrs/api/EncounterService.java#L432

omod/src/main/java/org/openmrs/web/controller/PortletController.java
```diff
@@ -425,4 +465,30 @@ public ModelAndView handleRequest(HttpServletRequest request, HttpServletRespons
 	protected void populateModel(HttpServletRequest request, Map<String, Object> model) {
 	}
 	
+	private Integer getStartIndexParameter(HttpServletRequest request) {
+		try {
+			return Integer.parseInt(request.getParameter("startIndex"));
```
dkayiwa, line 470: Can we also do the StringUtils.isNotBlank check for cases where startIndex may not be supplied?

omod/src/main/java/org/openmrs/web/controller/PortletController.java
```diff
@@ -425,4 +465,30 @@ public ModelAndView handleRequest(HttpServletRequest request, HttpServletRespons
...
+		}
+		catch (NumberFormatException e) {
+			return 0; // Return first page if not specified
+		}
+	}
+	
+	private Integer getLimitParameter(HttpServletRequest request, AdministrationService as, String globalPropertyKey) {
+		try {
+			String limitParam = request.getParameter("limit");
+			if (limitParam != null) {
```
dkayiwa, line 480: Can we instead use StringUtils.isNotBlank?

omod/src/main/java/org/openmrs/web/controller/PortletController.java
ODORA0: Changed

omod/src/main/java/org/openmrs/web/controller/PortletController.java
ODORA0: We storing `limit` in the model to make it accessible to/for the JSP and also keeping the pagination logic synchronised between the backend and JSP frontend

omod/src/main/java/org/openmrs/web/controller/PortletController.java
```diff
@@ -195,43 +204,68 @@ public ModelAndView handleRequest(HttpServletRequest request, HttpServletRespons
...
 							if (StringUtils.hasLength(weightString)) {
 								weightConcept = cs.getConceptNumeric(GeneralUtils.getConcept(weightString).getConceptId());
 							}
+							
 							String heightString = as.getGlobalProperty("concept.height");
 							ConceptNumeric heightConcept = null;
 							if (StringUtils.hasLength(heightString)) {
 								heightConcept = cs.getConceptNumeric(GeneralUtils.getConcept(heightString).getConceptId());
 							}
-							for (Obs obs : patientObs) {
```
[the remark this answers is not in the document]
ODORA0: Rather than loading all observations and finding the latest in memory, we would leverage the database to do this work more efficiently. Not directly related to pagination but aligns with the optimization goals since we no longer need to iterate over all patientObs to extract weight and height.

omod/src/main/java/org/openmrs/web/controller/PortletController.java
ODORA0: The [patientEncounters](https://github.com/openmrs/openmrs-module-legacyui/blob/89cdd965a1acf1c04909186edd78950a19fae668/omod/src/main/webapp/portlets/patientEncounters.jsp#L4) jsp

dkayiwa: Do you mind pointing me to the actual line that uses limit?

omod/src/main/java/org/openmrs/web/controller/PortletController.java
```diff
@@ -195,43 +204,68 @@ public ModelAndView handleRequest(HttpServletRequest request, HttpServletRespons
...
+								List<Obs> heightObs = Context.getObsService().getObservations(persons, null,
+								    heightQuestions, null, null, null, heightSort, 1, null, null, null, false);
+								
+								if (!heightObs.isEmpty()) {
+									latestHeight = heightObs.get(0);
+								}
 							}
+							
 							if (latestWeight != null && latestHeight != null) {
+								model.put("patientWeight", latestWeight);
```
dkayiwa, line 263: What happens to the model when latestHeight is null but latestWeight is not null?

omod/src/main/java/org/openmrs/web/controller/PortletController.java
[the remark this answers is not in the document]
dkayiwa: With performance optimisations, unless you have tested and confirmed, you can be very shocked to find a different outcome. The database itself being an expensive resource, and you have anyway already fetched these observations in the above call, you can find that actually iterating through the collection in memory may be faster that making another expensive database call. So i would only change that code if i have done some confirmations.

omod/src/main/java/org/openmrs/web/controller/PortletController.java
ODORA0: No BMI in the Model: Since both latestWeight and latestHeight are required for BMI calculation, the model will not have patientBmi or patientBmiAsString if either is null. he model will only include the observation that is not null (patientWeight in this case). But since I will revert this, I think any changes like setting a default value or log warning wouldn't be necessary

omod/src/main/java/org/openmrs/web/controller/PortletController.java
ODORA0: Looking at the dashboard JSP file, I don't actually see any direct reference to a limit or pagination for observations, safe to assume removing it. If the view needs the page size, it can derive it from patientObs.size()

omod/src/main/java/org/openmrs/web/controller/PortletController.java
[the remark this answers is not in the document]
ODORA0: @dkayiwa The paginated list only contains a limited number of observations (default 50), therefore BMI calculation couldn't find the necessary values, resulting in "?" if at all we add back the for loop while paginating. Now for this to work I would have to do something like this

```java
// Paginated observations for display
List<Obs> paginatedObs = Context.getObsService().getObservations(..., limit, startIndex, ...);

// Additional query to get ALL observations for BMI calculation
List<Obs> allObs = Context.getObsService().getObservations(persons, null, null, null, null,
    null, null, null, null, null, null, false);

// Finding weight/height in complete set of observations
for (Obs obs : allObs) {
    // ... looking for weight/height
}
``` 

In essence, we'd now be using:
- Paginated query for display purposes (performance optimization)
- Complete query for BMI calculation (functionality requirement)

Does this sound an approach to use?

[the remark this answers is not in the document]
dkayiwa: What do you see as the best approach?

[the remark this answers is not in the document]
ODORA0: I think I will go with this option I have suggested, in the long run we have separation of concerns. Change made

omod/src/main/java/org/openmrs/web/controller/PortletController.java
```diff
@@ -195,23 +204,42 @@ public ModelAndView handleRequest(HttpServletRequest request, HttpServletRespons
...
+						
+						Person person = (Person) p;
+						List<Person> persons = Collections.singletonList(person);
+						
+						// Get paginated observations
+						List<Obs> paginatedObs = Context.getObsService().getObservations(persons, null, null, null, null,
+						    null, Collections.singletonList("obsDatetime desc"), limit, startIndex, null, null, false);
+						
+						// Get all observations for BMI calculation
+						List<Obs> allObs = Context.getObsService().getObservations(persons, null, null, null, null, null,
```
dkayiwa, line 218: If you are loading all the patient's observations from the database into memory, then what optimisation is pagination doing?

ODORA0: Enhanced so that:
- Retrieves the latest observation for weight and height for the patient using the same observation retrieval method (getObservations).
- Sorting by obsDatetime desc ensures that only the most recent data is considered.
This has been tested and actually is optimised, loads faster for a patient with > 61k observation_count

omod/src/main/java/org/openmrs/web/controller/PortletController.java
```diff
@@ -195,43 +204,55 @@ public ModelAndView handleRequest(HttpServletRequest request, HttpServletRespons
...
-								            latestHeight.getObsDatetime()) > 0)) {
-									latestHeight = obs;
+								List<Obs> heightObs = Context.getObsService().getObservations(persons, null,
+								    Collections.singletonList(heightConcept), null, null, null,
+								    Collections.singletonList("obsDatetime desc"), 1, null, null, null, false);
+								if (!heightObs.isEmpty()) {
+									latestHeight = heightObs.get(0);
 								}
 							}
-							if (latestWeight != null) {
```
dkayiwa, line 226: According to the original code, latestHeight or latestWeight was put in the model as long as it was not null. Did you intentionally change it?

dkayiwa: Did you see the above comment?

PR #210, opened 2024-11-16
Attempt to restrict XSS filtering

ibacher:
Tries to limit the application of the XSS filter.


dkayiwa:
Will this prevent someone from submitting stored XSS via our REST calls?

ibacher:
In so far as they work on form parameters, it will. We need a slightly different solution for the REST API, though, as HTML-encoding will break anything where we expect the client to submit XML since, e.g.,

```
<?xml version="1.0"?>
<document />
```

Becomes

```
&lt;?xml version="1.0"?&gt;
&lt;document /&gt;
```

which won’t be parsed correctly. Basically for REST, we need to HTML encode the values as they are deserialised.

PR #219, opened 2025-08-06
LU-202: Migrate to OpenMRS Contrib DWR

wikumChamith:
Ticket: https://openmrs.atlassian.net/browse/LUI-202

### DIFF
--- a/omod/pom.xml
+++ b/omod/pom.xml
@@ -109,9 +109,9 @@
 		<!-- End OpenMRS core -->
 		
 		<dependency>
-			<groupId>org.openmrs.directwebremoting</groupId>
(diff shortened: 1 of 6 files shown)


omod/pom.xml
```diff
@@ -109,9 +109,9 @@
 		<!-- End OpenMRS core -->
 		
 		<dependency>
-			<groupId>org.openmrs.directwebremoting</groupId>
+			<groupId>org.openmrs.contrib</groupId>
```
dkayiwa, line 112: Is there a reason why you changed the group from `directwebremoting` to `contrib`?

wikumChamith: @ibacher suggested I do this to maintain consistency with our other projects.

PR #260, opened 2026-04-27
Adapt legacyui tests to TRUNK-6558 (JobRunr scheduler) and AddressSupport singleton

wikumChamith:
Two openmrs-core 3.0.0-SNAPSHOT changes broke the legacyui suite:                                                                                  
                            
  - TRUNK-6558 swapped the timer-based SchedulerService for JobRunr. TaskDefinition.taskInstance is no longer populated, so anything reading         
  task.getTaskInstance().isExecuting() is broken. Live state is now on SchedulerService.getTask(uuid).getState().
  - LUI-199 moved address-template reads behind the AddressSupport singleton, which caches per-JVM and inherits dirty state from earlier tests in the
   run.                                                                                                                                              
   
  Changes                                                                                                                                            
                            
  - SchedulerFormController — fix the "don't reschedule a running task" guard (was always true under JobRunr, so the controller silently rescheduled 
  in-flight jobs). Now uses getTask(uuid).getState() != PROCESSING.
  - ShortPatientFormValidatorTest / ShortPatientFormControllerTest — `@BeforeEach` resets the AddressSupport singleton: writes a clean                 
  layout.address.format GP and calls AddressSupport.getInstance().globalPropertyChanged(...) directly (the listener can be registered against a stale
   Spring context on CI).
  - TaskHelper.waitUntilTaskIsExecuting — polls SchedulerService.getTask(uuid).getState() instead of taskInstance. Wait budget bumped 2s→60s to match
   JobRunr's poll cadence.                                                                                                                           
  - TaskHelperTest.xml — repeat_interval 3600→0 so the scheduled job is a one-off and observable via getTask(task.uuid).
  - TaskHelperTest — both wait tests re-enabled.                                                                                                     
  - SchedulerFormControllerTest — the two reschedule tests stay `@Disabled`. Verifying them needs JobRunr to actually persist + run jobs across a test 
  transaction, which BaseModuleWebContextSensitiveTest's rollback-per-test base doesn't support. Equivalent coverage lives in core's                 
  SchedulerServiceIT.

### DIFF
--- a/omod/src/main/java/org/openmrs/scheduler/web/controller/SchedulerFormController.java
+++ b/omod/src/main/java/org/openmrs/scheduler/web/controller/SchedulerFormController.java
@@ -24,6 +24,8 @@ import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
 import org.openmrs.api.context.Context;
 import org.openmrs.scheduler.TaskDefinition;
+import org.openmrs.scheduler.TaskDetails;
+import org.openmrs.scheduler.TaskState;
 import org.openmrs.util.OpenmrsUtil;
 import org.openmrs.web.WebConstants;
 import org.openmrs.web.WebUtil;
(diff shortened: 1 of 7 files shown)


dkayiwa:
Did you get a chance to look at the failing builds?

wikumChamith:
> Did you get a chance to look at the failing builds?

Yeah I am looking at them. I am only seeing them on GHA

wikumChamith:
> Did you get a chance to look at the failing builds?

Fixed it :)

dkayiwa:
@coderabbitai full review

PR #262, opened 2026-05-04
Implement authorization checks for all controllers and DWR endpoints

ibacher:


dkayiwa:
Is the the `@RequirePrivilege` annotation required for controllers that already call the service layer methods which require privileges?

ibacher:
> Is the the @RequirePrivilege annotation required for controllers that already call the service layer methods which require privileges?

For controllers that extend the `SimpleFormController`, there are some circumstances where it's possible for the changes to get written even when the service layer rejects the request, so it's a good idea.

dkayiwa:
Oh i see!!! 👍

omod/src/main/resources/config.xml
```diff
@@ -177,15 +176,10 @@
...
 				<param name="class" value="org.openmrs.web.dwr.DWRRelationshipService"/>
-				<include method="getRelationshipTypes"/>
-				<include method="getRelationshipTypeId"/>
 				<include method="createRelationship"/>
 				<include method="voidRelationship"/>
 				<include method="changeRelationshipDates"/>
-				<include method="setRelationshipTo"/>
 				<include method="getRelationships"/>
-				<include method="getRelationshipsToPerson"/>
-				<include method="getRelationshipsFromPerson"/>
```
mseaton, line 188: Are we sure these removed methods are not used by anything downstream?

ibacher: If they are, the downstream already errors out as the underlying methods don't exist on the DWR services, which is the only reason they're being removed.

PR #265, opened 2026-06-16
O3-5739: Skip XSS request wrapping for web-service paths

wikumChamith:
XSSFilter no longer wraps requests whose URI contains /ws/, so REST and FHIR request bodies reach the resource layer as raw bytes.

https://openmrs.atlassian.net/browse/O3-5739

### DIFF
--- a/omod/src/main/java/org/openmrs/web/xss/XSSFilter.java
+++ b/omod/src/main/java/org/openmrs/web/xss/XSSFilter.java
@@ -23,19 +23,20 @@ import org.apache.commons.fileupload.servlet.ServletFileUpload;
 import org.springframework.web.multipart.support.DefaultMultipartHttpServletRequest;
 
 public class XSSFilter implements Filter {
-	
+
+	private static final String WEB_SERVICE_PATH = "/ws/";
+
 	@Override
 	public void doFilter(ServletRequest request, ServletResponse response, FilterChain chain) throws IOException,
 	        ServletException {
-		
-		if (!"GET".equalsIgnoreCase(((HttpServletRequest) request).getMethod())) {
-			if (ServletFileUpload.isMultipartContent((HttpServletRequest) request)) {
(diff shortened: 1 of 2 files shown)


wikumChamith:
> ## [Quality Gate Failed](https://sonarcloud.io/dashboard?id=openmrs_openmrs-module-legacyui&pullRequest=265) **Quality Gate failed**
> 
> Failed conditions  [B Maintainability Rating on New Code](https://sonarcloud.io/dashboard?id=openmrs_openmrs-module-legacyui&pullRequest=265) (required ≥ A)
> 
> [See analysis details on SonarQube Cloud](https://sonarcloud.io/dashboard?id=openmrs_openmrs-module-legacyui&pullRequest=265)
> 
>  Catch issues before they fail your Quality Gate with our IDE extension  [SonarQube for IDE](https://www.sonarsource.com/products/sonarlint/features/connected-mode/?referrer=pull-request)

This feels like an overreach:	https://sonarcloud.io/project/issues?sinceLeakPeriod=true&issueStatuses=OPEN%2CCONFIRMED&pullRequest=265&id=openmrs_openmrs-module-legacyui&open=AZ7OwGPqhsNhAUEFGDCj

ibacher:
So, while I do think the `XSSFilter` is the cause of the problem, I don't think that just dropping requests to `/ws/` winds up being a complete or even _good_ solution. Basically this just creates an asymmetry between "data entered via Java forms" and "data entered via the REST API", but worse, it makes the REST API a backdoor for loading XSS-style strings that are then incorrectly displayed in the legacy or O2 UI, basically removing the whole reason for having this filter in the first place.

Ultimately, rather than this being something we can just hack-up a quick patch around, I think we need to come up with a principled way to properly re-encode output at the point it's displayed.

wikumChamith:
So isn't it better to fix this in the frontend?

ibacher:
> So isn't it better to fix this in the frontend?

Yep. And probably also the legacy UI...

PR #267, opened 2026-07-19
LUI-213: Support viewing archived observations in Legacy UI

shivvani-r:
## Description of what I changed
This PR updates the Legacy UI to properly fetch and display archived observations, ensuring they are not lost to the user after being archived - implemented as part of https://openmrs.atlassian.net/browse/TRUNK-6648.

## Specific changes include

- `DWRObsService.java`: Merged archived observations into the search results so they can be viewed and managed on the "Manage Observations" page.
- `EncounterFormController.java`: Switched to `getAllObsIncludingArchived()` to ensure archived observations are passed to the view, restoring the functionality of the "Toggle Voided" button on the Encounter form.
- `obs.js`: Updated the DWR rowCreator to correctly apply the .voided CSS class so that voided observations dynamically render with the standard OpenMRS strikethrough styling while preserving alternating row colors.

## Issue I worked on
see https://openmrs.atlassian.net/browse/LUI-213

### DIFF
--- a/omod/src/main/java/org/openmrs/web/controller/encounter/EncounterFormController.java
+++ b/omod/src/main/java/org/openmrs/web/controller/encounter/EncounterFormController.java
@@ -315,7 +315,7 @@ public class EncounterFormController extends SimpleFormController {
 			map.put("encounterRoles", es.getAllEncounterRoles(false));
 			map.put("forms", Context.getFormService().getAllForms());
 			// loop over the encounter's observations to find the edited obs
-			for (Obs o : encounter.getObsAtTopLevel(true)) {
+			for (Obs o : encounter.getAllObsIncludingArchived()) {
 				
 				// only edited obs has previous version
 				if (o.hasPreviousVersion()) {
--- a/omod/src/main/java/org/openmrs/web/controller/encounter/EncounterFormController.java
+++ b/omod/src/main/java/org/openmrs/web/controller/encounter/EncounterFormController.java
@@ -295,6 +295,7 @@ public class EncounterFormController extends SimpleFormController {
 			obsMapToReturn = new TreeMap<FormField, List<Obs>>(new NumberingFormFieldComparator()); // use custom comparator
 		}
 		
+
 		if (Context.isAuthenticated()) {
 			EncounterService es = Context.getEncounterService();
 			FormService fs = Context.getFormService();
@@ -320,9 +321,14 @@ public class EncounterFormController extends SimpleFormController {
 					List<Obs> list = groupMembersMap.get(parent);
 					if (list == null) {
 						list = new ArrayList<Obs>();
+						if (parent.hasGroupMembers()) {
+							list.addAll(parent.getGroupMembers(true));
+						}
 						groupMembersMap.put(parent, list);
 					}
-					list.add(o);
+					if (!list.contains(o)) {
+						list.add(o);
+					}
 					continue;
 				}
 				
@@ -349,6 +355,7 @@ public class EncounterFormController extends SimpleFormController {
 		}
 		
 		if (log.isDebugEnabled()) {
+			log.error("Total obs: " + encounter.getAllObsIncludingArchived().size());
 			log.debug("setting obsMap in page context (size: " + obsMapToReturn.size() + ")");
 		}
 		map.put("obsMap", obsMapToReturn);
--- a/omod/src/main/java/org/openmrs/web/controller/encounter/EncounterFormController.java
+++ b/omod/src/main/java/org/openmrs/web/controller/encounter/EncounterFormController.java
@@ -355,7 +355,6 @@ public class EncounterFormController extends SimpleFormController {
 		}
 		
 		if (log.isDebugEnabled()) {
-			log.error("Total obs: " + encounter.getAllObsIncludingArchived().size());
 			log.debug("setting obsMap in page context (size: " + obsMapToReturn.size() + ")");
 		}
 		map.put("obsMap", obsMapToReturn);
--- a/omod/src/main/java/org/openmrs/web/controller/encounter/EncounterFormController.java
+++ b/omod/src/main/java/org/openmrs/web/controller/encounter/EncounterFormController.java
@@ -295,11 +295,6 @@ public class EncounterFormController extends SimpleFormController {
 			obsMapToReturn = new TreeMap<FormField, List<Obs>>(new NumberingFormFieldComparator()); // use custom comparator
 		}
 		
-		// this maps the obs to form field objects for non top-level obs
-		// it is keyed on obs so that when looping over an exploded obsGroup
-		// the formfield can be fetched easily (in order to show the field numbers etc)
-		Map<Obs, FormField> otherFormFields = new HashMap<Obs, FormField>();
-		
 		if (Context.isAuthenticated()) {
 			EncounterService es = Context.getEncounterService();
 			FormService fs = Context.getFormService();
@@ -357,9 +352,6 @@ public class EncounterFormController extends SimpleFormController {
 			log.debug("setting obsMap in page context (size: " + obsMapToReturn.size() + ")");
 		}
 		map.put("obsMap", obsMapToReturn);
-		
-		map.put("otherFormFields", otherFormFields);
-		
 		map.put("locale", Context.getLocale());
 		map.put("editedObs", editedObs);
 		if (encounter.getPatient() != null) {
--- a/omod/src/main/java/org/openmrs/web/controller/encounter/EncounterFormController.java
+++ b/omod/src/main/java/org/openmrs/web/controller/encounter/EncounterFormController.java
@@ -314,8 +314,22 @@ public class EncounterFormController extends SimpleFormController {
 			
 			map.put("encounterRoles", es.getAllEncounterRoles(false));
 			map.put("forms", Context.getFormService().getAllForms());
+			
+			Map<Obs, List<Obs>> groupMembersMap = new HashMap<Obs, List<Obs>>();
+			
 			// loop over the encounter's observations to find the edited obs
 			for (Obs o : encounter.getAllObsIncludingArchived()) {
+				
+				Obs parent = o.getObsGroup();
+				if (parent != null) {
+					List<Obs> list = groupMembersMap.get(parent);
+					if (list == null) {
+						list = new ArrayList<Obs>();
+						groupMembersMap.put(parent, list);
+					}
+					list.add(o);
+					continue;
+				}
 				
 				// only edited obs has previous version
 				if (o.hasPreviousVersion()) {
@@ -328,22 +342,15 @@ public class EncounterFormController extends SimpleFormController {
 					ff = new FormField();
 				}
 				
-				// we only put the top-level obs in the obsMap.  Those would
-				// be the obs that don't have an obs grouper 
-				if (o.getObsGroup() == null) {
-					// populate the obs map with this formfield and obs
-					List<Obs> list = obsMapToReturn.get(ff);
-					if (list == null) {
-						list = new Vector<Obs>();
-						obsMapToReturn.put(ff, list);
-					}
-					list.add(o);
-				} else {
-					// this is not a top-level obs, just put the formField
-					// in a separate list and be done with it
-					otherFormFields.put(o, ff);
+				// populate the obs map with this formfield and obs
+				List<Obs> list = obsMapToReturn.get(ff);
+				if (list == null) {
+					list = new Vector<Obs>();
+					obsMapToReturn.put(ff, list);
 				}
+				list.add(o);
 			}
+			map.put("groupMembersMap", groupMembersMap);
 		}
 		
 		if (log.isDebugEnabled()) {
--- a/omod/src/main/java/org/openmrs/web/dwr/DWRObsService.java
+++ b/omod/src/main/java/org/openmrs/web/dwr/DWRObsService.java
@@ -11,7 +11,9 @@ package org.openmrs.web.dwr;
 
 import java.text.ParseException;
 import java.text.SimpleDateFormat;
+import java.util.ArrayList;
 import java.util.Collection;
+import java.util.Collections;
 import java.util.Date;
 import java.util.Set;
 import java.util.Vector;
@@ -29,6 +31,7 @@ import org.openmrs.Obs;
 import org.openmrs.Person;
 import org.openmrs.api.AdministrationService;
 import org.openmrs.api.context.Context;
+import org.openmrs.api.impl.ObsArchiveHelper;
 import org.openmrs.util.OpenmrsUtil;
 import org.openmrs.util.PrivilegeConstants;
 import org.openmrs.web.security.RequirePrivilege;
@@ -317,13 +320,23 @@ public class DWRObsService {
 		
 		if (p != null && c != null) {
 			log.debug("Getting obss with patient and concept");
-			obss = Context.getObsService().getObservationsByPersonAndConcept(p, c);
+			obss = new ArrayList<Obs>(Context.getObsService().getObservations(
+			    Collections.singletonList(p), null, Collections.singletonList(c), null, null, null, null, null, null, null, null, true));
+			ObsArchiveHelper archiveHelper = Context.getRegisteredComponent("obsArchiveHelper", ObsArchiveHelper.class);
+			if (archiveHelper != null) {
+				obss.addAll(archiveHelper.getArchivedObsByPersonIdAndConceptId(p.getPersonId(), c.getConceptId()));
+			}
 		} else if (e != null) {
 			log.debug("Getting obss by encounter");
-			obss = e.getAllObs();
+			obss = e.getAllObsIncludingArchived();
 		} else if (p != null) {
 			log.debug("Getting obss with just patient");
-			obss = Context.getObsService().getObservationsByPerson(p);
+			obss = new ArrayList<Obs>(Context.getObsService().getObservations(
+			    Collections.singletonList(p), null, null, null, null, null, null, null, null, null, null, true));
+			ObsArchiveHelper archiveHelper = Context.getRegisteredComponent("obsArchiveHelper", ObsArchiveHelper.class);
+			if (archiveHelper != null) {
+				obss.addAll(archiveHelper.getArchivedObsByPersonId(p.getPersonId()));
+			}
 		}
 		
 		if (obss != null) {
--- a/omod/src/main/java/org/openmrs/web/dwr/DWRObsService.java
+++ b/omod/src/main/java/org/openmrs/web/dwr/DWRObsService.java
@@ -35,7 +35,6 @@ import org.openmrs.api.AdministrationService;
 import org.openmrs.api.APIException;
 import org.openmrs.api.context.Context;
 import org.openmrs.api.impl.ObsArchiveHelper;
-import org.openmrs.util.OpenmrsConstants;
 import org.openmrs.util.OpenmrsUtil;
 import org.openmrs.util.PrivilegeConstants;
 import org.openmrs.web.security.RequirePrivilege;
@@ -330,7 +329,7 @@ public class DWRObsService {
 			e = Context.getEncounterService().getEncounter(eId);
 		}
 		
-		Collection<Obs> obss = null;
+		List<Obs> obss = null;
 		
 		if (p != null && c != null) {
 			log.debug("Getting obss with patient and concept");
@@ -340,11 +339,11 @@ public class DWRObsService {
 			if (archiveHelper != null) {
 				obss.addAll(archiveHelper.getArchivedObsByPersonIdAndConceptId(p.getPersonId(), c.getConceptId()));
 			}
-			sortObservations((List<Obs>) obss);
+			sortObservations(obss);
 		} else if (e != null) {
 			log.debug("Getting obss by encounter");
 			obss = new ArrayList<Obs>(e.getAllObsIncludingArchived());
-			sortObservations((List<Obs>) obss);
+			sortObservations(obss);
 		} else if (p != null) {
 			log.debug("Getting obss with just patient");
 			obss = new ArrayList<Obs>(Context.getObsService().getObservations(
@@ -353,7 +352,7 @@ public class DWRObsService {
 			if (archiveHelper != null) {
 				obss.addAll(archiveHelper.getArchivedObsByPersonId(p.getPersonId()));
 			}
-			sortObservations((List<Obs>) obss);
+			sortObservations(obss);
 		}
 		
 		if (obss != null) {
--- a/omod/src/main/java/org/openmrs/web/dwr/DWRObsService.java
+++ b/omod/src/main/java/org/openmrs/web/dwr/DWRObsService.java
@@ -30,8 +30,10 @@ import org.openmrs.Location;
 import org.openmrs.Obs;
 import org.openmrs.Person;
 import org.openmrs.api.AdministrationService;
+import org.openmrs.api.APIException;
 import org.openmrs.api.context.Context;
 import org.openmrs.api.impl.ObsArchiveHelper;
+import org.openmrs.util.OpenmrsConstants;
 import org.openmrs.util.OpenmrsUtil;
 import org.openmrs.util.PrivilegeConstants;
 import org.openmrs.web.security.RequirePrivilege;
@@ -93,6 +95,20 @@ public class DWRObsService {
 		}
 		
 		return obsList;
+	}
+
+	private ObsArchiveHelper getObsArchiveHelperSafely() {
+		try {
+			String lastProcessedId = Context.getAdministrationService()
+			        .getGlobalProperty(OpenmrsConstants.GP_OBS_ARCHIVE_LAST_PROCESSED_OBS_ID);
+			if (lastProcessedId != null && !lastProcessedId.trim().isEmpty()) {
+				return Context.getRegisteredComponent("obsArchiveHelper", ObsArchiveHelper.class);
+			}
+		}
+		catch (APIException e) {
+			// archive table may not exist yet or context not available, degrade gracefully
+		}
+		return null;
 	}
 	
 	/**
@@ -322,7 +338,7 @@ public class DWRObsService {
 			log.debug("Getting obss with patient and concept");
 			obss = new ArrayList<Obs>(Context.getObsService().getObservations(
 			    Collections.singletonList(p), null, Collections.singletonList(c), null, null, null, null, null, null, null, null, true));
-			ObsArchiveHelper archiveHelper = Context.getRegisteredComponent("obsArchiveHelper", ObsArchiveHelper.class);
+			ObsArchiveHelper archiveHelper = getObsArchiveHelperSafely();
 			if (archiveHelper != null) {
 				obss.addAll(archiveHelper.getArchivedObsByPersonIdAndConceptId(p.getPersonId(), c.getConceptId()));
 			}
@@ -333,7 +349,7 @@ public class DWRObsService {
 			log.debug("Getting obss with just patient");
 			obss = new ArrayList<Obs>(Context.getObsService().getObservations(
 			    Collections.singletonList(p), null, null, null, null, null, null, null, null, null, null, true));
-			ObsArchiveHelper archiveHelper = Context.getRegisteredComponent("obsArchiveHelper", ObsArchiveHelper.class);
+			ObsArchiveHelper archiveHelper = getObsArchiveHelperSafely();
 			if (archiveHelper != null) {
 				obss.addAll(archiveHelper.getArchivedObsByPersonId(p.getPersonId()));
 			}
--- a/omod/src/main/java/org/openmrs/web/dwr/DWRObsService.java
+++ b/omod/src/main/java/org/openmrs/web/dwr/DWRObsService.java
@@ -101,14 +101,10 @@ public class DWRObsService {
 
 	private ObsArchiveHelper getObsArchiveHelperSafely() {
 		try {
-			String lastProcessedId = Context.getAdministrationService()
-			        .getGlobalProperty(OpenmrsConstants.GP_OBS_ARCHIVE_LAST_PROCESSED_OBS_ID);
-			if (lastProcessedId != null && !lastProcessedId.trim().isEmpty()) {
-				return Context.getRegisteredComponent("obsArchiveHelper", ObsArchiveHelper.class);
-			}
+			return Context.getRegisteredComponent("obsArchiveHelper", ObsArchiveHelper.class);
 		}
 		catch (APIException e) {
-			// archive table may not exist yet or context not available, degrade gracefully
+			// bean not registered on this core, degrade gracefully
 		}
 		return null;
 	}
--- a/omod/src/main/java/org/openmrs/web/dwr/DWRObsService.java
+++ b/omod/src/main/java/org/openmrs/web/dwr/DWRObsService.java
@@ -342,6 +342,7 @@ public class DWRObsService {
 			if (archiveHelper != null) {
 				obss.addAll(archiveHelper.getArchivedObsByPersonIdAndConceptId(p.getPersonId(), c.getConceptId()));
 			}
+			sortObservations((java.util.List<Obs>) obss);
 		} else if (e != null) {
 			log.debug("Getting obss by encounter");
 			obss = e.getAllObsIncludingArchived();
@@ -353,6 +354,7 @@ public class DWRObsService {
 			if (archiveHelper != null) {
 				obss.addAll(archiveHelper.getArchivedObsByPersonId(p.getPersonId()));
 			}
+			sortObservations((java.util.List<Obs>) obss);
 		}
 		
 		if (obss != null) {
@@ -384,5 +386,18 @@ public class DWRObsService {
 		}
 		
 		return oItem;
+	}
+
+	private void sortObservations(java.util.List<Obs> obss) {
+		Collections.sort(obss, new java.util.Comparator<Obs>() {
+			@Override
+			public int compare(Obs o1, Obs o2) {
+				int result = OpenmrsUtil.compareWithNullAsEarliest(o2.getObsDatetime(), o1.getObsDatetime());
+				if (result == 0) {
+					return OpenmrsUtil.compareWithNullAsGreatest(o1.getObsId(), o2.getObsId());
+				}
+				return result;
+			}
+		});
 	}
 }
--- a/omod/src/main/java/org/openmrs/web/dwr/DWRObsService.java
+++ b/omod/src/main/java/org/openmrs/web/dwr/DWRObsService.java
@@ -14,7 +14,9 @@ import java.text.SimpleDateFormat;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
+import java.util.Comparator;
 import java.util.Date;
+import java.util.List;
 import java.util.Set;
 import java.util.Vector;
 
@@ -342,10 +344,11 @@ public class DWRObsService {
 			if (archiveHelper != null) {
 				obss.addAll(archiveHelper.getArchivedObsByPersonIdAndConceptId(p.getPersonId(), c.getConceptId()));
 			}
-			sortObservations((java.util.List<Obs>) obss);
+			sortObservations((List<Obs>) obss);
 		} else if (e != null) {
 			log.debug("Getting obss by encounter");
-			obss = e.getAllObsIncludingArchived();
+			obss = new ArrayList<Obs>(e.getAllObsIncludingArchived());
+			sortObservations((List<Obs>) obss);
 		} else if (p != null) {
 			log.debug("Getting obss with just patient");
 			obss = new ArrayList<Obs>(Context.getObsService().getObservations(
@@ -354,7 +357,7 @@ public class DWRObsService {
 			if (archiveHelper != null) {
 				obss.addAll(archiveHelper.getArchivedObsByPersonId(p.getPersonId()));
 			}
-			sortObservations((java.util.List<Obs>) obss);
+			sortObservations((List<Obs>) obss);
 		}
 		
 		if (obss != null) {
@@ -388,8 +391,8 @@ public class DWRObsService {
 		return oItem;
 	}
 
-	private void sortObservations(java.util.List<Obs> obss) {
-		Collections.sort(obss, new java.util.Comparator<Obs>() {
+	private void sortObservations(List<Obs> obss) {
+		Collections.sort(obss, new Comparator<Obs>() {
 			@Override
 			public int compare(Obs o1, Obs o2) {
 				int result = OpenmrsUtil.compareWithNullAsEarliest(o2.getObsDatetime(), o1.getObsDatetime());
--- a/omod/src/main/webapp/admin/encounters/obsDisplay.jsp
+++ b/omod/src/main/webapp/admin/encounters/obsDisplay.jsp
@@ -15,7 +15,7 @@
 			<tr class="<c:if test="${obs.voided}">voided </c:if>">
 				<td colspan="5"><%-- this is the empty row to mimic the description row--%></td>
 			</tr>
-			<c:set var="obsList" value="${obs.groupMembers}" scope="request"/>
+			<c:set var="obsList" value="${groupMembersMap != null && groupMembersMap[obs] != null ? groupMembersMap[obs] : obs.groupMembers}" scope="request"/>
 			<c:set var="field" value="${otherFormFields[groupMember]}" scope="request"/>
 			<c:set var="level" value="${level+1}" scope="request"/>
 			<c:import url="obsDisplay.jsp" />
--- a/omod/src/main/webapp/admin/encounters/obsDisplay.jsp
+++ b/omod/src/main/webapp/admin/encounters/obsDisplay.jsp
@@ -16,6 +16,7 @@
 				<td colspan="5"><%-- this is the empty row to mimic the description row--%></td>
 			</tr>
 			<c:set var="obsList" value="${groupMembersMap != null && groupMembersMap[obs] != null ? groupMembersMap[obs] : obs.groupMembers}" scope="request"/>
+			<c:remove var="field" scope="request"/>
 			<c:set var="level" value="${level+1}" scope="request"/>
 			<c:import url="obsDisplay.jsp" />
 		</c:when>
--- a/omod/src/main/webapp/admin/encounters/obsDisplay.jsp
+++ b/omod/src/main/webapp/admin/encounters/obsDisplay.jsp
@@ -7,7 +7,7 @@
 
 <c:forEach var="obs" items="${obsList}">
 	<c:choose>
-		<c:when test="${obs.obsGrouping}">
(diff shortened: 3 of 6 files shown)


omod/src/main/java/org/openmrs/web/dwr/DWRObsService.java
```diff
@@ -317,13 +320,23 @@ public Vector<ObsListItem> getObsByPatientConceptEncounter(String personId, Stri
 		
 		if (p != null && c != null) {
 			log.debug("Getting obss with patient and concept");
-			obss = Context.getObsService().getObservationsByPersonAndConcept(p, c);
+			obss = new ArrayList<Obs>(Context.getObsService().getObservations(
+			    Collections.singletonList(p), null, Collections.singletonList(c), null, null, null, null, null, null, null, null, true));
+			ObsArchiveHelper archiveHelper = Context.getRegisteredComponent("obsArchiveHelper", ObsArchiveHelper.class);
+			if (archiveHelper != null) {
+				obss.addAll(archiveHelper.getArchivedObsByPersonIdAndConceptId(p.getPersonId(), c.getConceptId()));
```
dkayiwa, line 340: This one can't go in yet: neither `ObsArchiveHelper` nor `Encounter.getAllObsIncludingArchived()` exists in any published openmrs-core 3.0.0-SNAPSHOT, which is why the build check on this PR fails with "cannot find symbol". If it were merged as-is, master here would stop compiling for everyone. Both symbols currently live only in the still-open core PR openmrs/openmrs-core#6171, and the two person-based lookups used here (`getArchivedObsByPersonIdAndConceptId` and `getArchivedObsByPersonId`) don't exist even in that PR; its helper only has `isArchived`, `hasArchivedChildren`, `getArchivedObsByEncounterId`, `getObsFromArchive`/`getObsFromArchiveByUuid` and `restoreFromArchive`. So the core side needs to land first (including adding these person-based methods there or in a follow-up), a SNAPSHOT containing them needs to be deployed, and CI here needs to go green before this can merge.

omod/src/main/java/org/openmrs/web/dwr/DWRObsService.java
```diff
@@ -317,13 +320,23 @@ public Vector<ObsListItem> getObsByPatientConceptEncounter(String personId, Stri
 		
 		if (p != null && c != null) {
 			log.debug("Getting obss with patient and concept");
-			obss = Context.getObsService().getObservationsByPersonAndConcept(p, c);
+			obss = new ArrayList<Obs>(Context.getObsService().getObservations(
+			    Collections.singletonList(p), null, Collections.singletonList(c), null, null, null, null, null, null, null, null, true));
+			ObsArchiveHelper archiveHelper = Context.getRegisteredComponent("obsArchiveHelper", ObsArchiveHelper.class);
```
dkayiwa: The `archiveHelper != null` guards (this branch and the person-only one below) can never be false. `Context.getRegisteredComponent` throws `APIException` when the bean isn't registered; it never returns null. So on a core without the helper bean this doesn't quietly skip archived results, it fails the whole search request with a DWR error, which I suspect is the opposite of what the guard is trying to do. Core handles this exact situation in `Encounter.getAllObsIncludingArchived()` by wrapping the lookup in try/catch (and checking the `obs.archive.last_processed_obs_id` global property before querying). I'd mirror that pattern here, e.g. a small private method that returns the helper or null, so the null checks actually mean something. Not a blocker on its own, since once the core PR lands the bean is registered unconditionally.

omod/src/main/java/org/openmrs/web/dwr/DWRObsService.java
```diff
@@ -317,13 +320,23 @@ public Vector<ObsListItem> getObsByPatientConceptEncounter(String personId, Stri
 		
 		if (p != null && c != null) {
 			log.debug("Getting obss with patient and concept");
-			obss = Context.getObsService().getObservationsByPersonAndConcept(p, c);
+			obss = new ArrayList<Obs>(Context.getObsService().getObservations(
```
dkayiwa, line 336: When this gets re-spun after the core API lands, please add a test that pins the new behavior. Nothing exercises this path right now (Sonar reports 0% coverage on the new code, and `DWRObservationServiceTest` only covers `createObs`). A `BaseModuleWebContextSensitiveTest` case that seeds one voided obs plus one row in `obs_archive` (core's `ObsArchiveIntegrationTest` shows the fixture pattern) and asserts that `getObsByPatientConceptEncounter` returns both, with `voided=true` on the resulting `ObsListItem`s, would cover both the merge logic and the DWR-visible flag the new obs.js styling depends on.

omod/src/main/java/org/openmrs/web/controller/encounter/EncounterFormController.java
```diff
@@ -315,7 +315,7 @@ protected Map<String, Object> referenceData(HttpServletRequest request, Object o
 			map.put("encounterRoles", es.getAllEncounterRoles(false));
 			map.put("forms", Context.getFormService().getAllForms());
 			// loop over the encounter's observations to find the edited obs
-			for (Obs o : encounter.getObsAtTopLevel(true)) {
+			for (Obs o : encounter.getAllObsIncludingArchived()) {
```
dkayiwa, line 317: Was the wider iteration here intended, beyond pulling in archived obs? Two side effects I'd like to confirm: (1) `getObsAtTopLevel(true)` only walked top-level obs, while `getAllObsIncludingArchived()` walks every obs on the encounter, so group members now go through this loop too. That means `editedObs` can now contain member ids, and `getFormField` runs once per member on every page render, so a large grouped encounter does quite a few more lookups than before. (2) This surfaces archived top-level obs only: obsDisplay.jsp renders group members from `obs.groupMembers`, which reads the obs table, and core PR openmrs/openmrs-core#6171 doesn't make that relation archive-aware, so an archived group member still won't show up under Toggle Voided. If grouped obs are out of scope for LUI-213 that's fine, but it would be worth saying so on the ticket.

omod/src/main/java/org/openmrs/web/dwr/DWRObsService.java
shivvani-r: Added both getArchivedObsByPersonId and getArchivedObsByPersonIdAndConceptId to the ObsArchiveHelper. Forgot to push them.

omod/src/main/java/org/openmrs/web/dwr/DWRObsService.java
shivvani-r: I've refactored it to mirror the pattern used in Encounter.java

omod/src/main/java/org/openmrs/web/dwr/DWRObsService.java
shivvani-r: I've added an integration test (getObsByPatientConceptEncounter_shouldIncludeArchivedObs) to DWRObservationServiceTest.java

omod/src/main/java/org/openmrs/web/controller/encounter/EncounterFormController.java
shivvani-r: thanks for pointing this out 😅 
I kept the wider iteration to access the children, but I added a continue condition to ensure we completely skip running getFormField for any child observations (since they don't have form fields).

Since obs.groupMembers doesnt know about the archive table, I updated EncounterFormController.java to put all child observations (including the archived ones) into a groupMembersMap. Archived children are now visible in both the searches-- toggle deleted and under manage observations

omod/src/main/java/org/openmrs/web/controller/encounter/EncounterFormController.java
```diff
@@ -328,22 +342,15 @@ protected Map<String, Object> referenceData(HttpServletRequest request, Object o
...
-					List<Obs> list = obsMapToReturn.get(ff);
-					if (list == null) {
-						list = new Vector<Obs>();
-						obsMapToReturn.put(ff, list);
-					}
-					list.add(o);
-				} else {
-					// this is not a top-level obs, just put the formField
-					// in a separate list and be done with it
-					otherFormFields.put(o, ff);
```
dkayiwa, line 344: Worth deleting the leftovers here: dropping this `else` leaves `otherFormFields` with no writer at all, so the map still published further down (line 361 on your branch) is now always empty. I confirmed that after `referenceData` runs, it comes back with size 0.

Nothing in legacyui depends on it, so there is no breakage. obsDisplay.jsp's only read is `${otherFormFields[groupMember]}`, and `groupMember` has never been a defined variable in that page, so the lookup already resolved to null. But the local at line 301 and the `map.put` are dead weight now, and dropping them saves the next reader hunting for what is supposed to fill them.

omod/src/main/java/org/openmrs/web/dwr/DWRObsService.java
dkayiwa: openmrs/openmrs-core#6171 is still open, so the published 3.0.0-SNAPSHOT carries none of these symbols and the build here is still failing on the same eight `cannot find symbol` errors (`ObsArchiveHelper`, `Encounter.getAllObsIncludingArchived()`, `OpenmrsConstants.GP_OBS_ARCHIVE_LAST_PROCESSED_OBS_ID`).

To get past that I built this branch locally against your core branch (3921b237), and it does compile there and the omod tests pass, so ordering is all that is left on this front: core merges, a SNAPSHOT carrying it gets deployed, then CI here goes green.

omod/src/main/java/org/openmrs/web/controller/encounter/EncounterFormController.java
dkayiwa: `groupMembersMap` only gets read when the parent still has at least one member row left in `obs`. obsDisplay.jsp consults it inside `<c:when test="${obs.obsGrouping}">`, and `isObsGrouping()` is `hasGroupMembers(true)` over the live `groupMembers` collection, which `ObsArchiveHelper.convertToObs` never populates. So once every member of a group has been archived, the parent falls through to `<c:otherwise>`, renders as a single leaf row, and the members `referenceData` just collected are never rendered at all.

I ran this as a controller test against your core branch. Seeding encounter 3 with an archived parent (998) and its archived member (999):

```
obsMap top level:  obsId=998  isObsGrouping=false   -> recursion SKIPPED
groupMembersMap:   parent=998 members=[999]
```

Same outcome for a still-live parent whose only member had been archived. A parent that keeps one live member does work: `parent=997 members=[996 live, 995 archived]` with `isObsGrouping=true`, and both render.

That is exactly the shape the archiving task produces, since it holds a parent back until no active children are left in `obs` and then archives it together with them. So the members of a fully archived group stay invisible on the encounter form, and nothing throws, the rows are simply missing.

Gating the recursion on the map as well as the live collection is enough, on line 10 of obsDisplay.jsp:

```jsp
<c:when test="${obs.obsGrouping || (groupMembersMap != null && groupMembersMap[obs] != null)}">
```

If you would rather leave fully archived groups out of LUI-213, that is a reasonable call, but please say so on the ticket rather than leaving it looking covered.

Either way this path wants a test. `EncounterFormControllerTest` already exists; a case that seeds an archived group and asserts the member is reachable from what the JSP iterates would have caught this.

omod/src/main/java/org/openmrs/web/dwr/DWRObsService.java
shivvani-r: can we rerun the CI/CD once

omod/src/main/java/org/openmrs/web/controller/encounter/EncounterFormController.java
shivvani-r: ive removed it!

omod/src/main/java/org/openmrs/web/controller/encounter/EncounterFormController.java
shivvani-r: Thanks I've applied the fix you have suggested. I also added the integration test to EncounterFormControllerTest.

omod/src/main/webapp/admin/encounters/obsDisplay.jsp
```diff
@@ -7,16 +7,15 @@
...
 			<tr class="<c:if test="${obs.voided}">voided </c:if>obsGroupHeader">
 				<td>${field.fieldNumber}<c:if test="${field.fieldPart != null && field.fieldPart != ''}">.${field.fieldPart}</c:if></td>
 				<td colspan="4" style="padding-left: ${padding}"><c:out value="${field.field.concept.name.name}" /></td>
 			</tr>
 			<tr class="<c:if test="${obs.voided}">voided </c:if>">
 				<td colspan="5"><%-- this is the empty row to mimic the description row--%></td>
 			</tr>
-			<c:set var="obsList" value="${obs.groupMembers}" scope="request"/>
-			<c:set var="field" value="${otherFormFields[groupMember]}" scope="request"/>
+			<c:set var="obsList" value="${groupMembersMap != null && groupMembersMap[obs] != null ? groupMembersMap[obs] : obs.groupMembers}" scope="request"/>
```
dkayiwa, line 18: Deleting the old `<c:set var="field" value="${otherFormFields[groupMember]}" scope="request"/>` from below this line changes what group members render, was that intended? `otherFormFields[groupMember]` was always null (`groupMember` is not a defined variable anywhere in the page), and JSTL removes the scoped attribute when the value is null (`SetSupport.doEndTag` calls `pageContext.removeAttribute(var, scope)`), so that line was clearing request-scoped `field` before the recursion and member rows came out with an empty field-number column. With it gone, the recursion inherits the parent's `FormField` from `encounterForm.jsp:551`, so every member row now prints the group's field number, and a nested group header prints its ancestor's concept name in `${field.field.concept.name.name}`.

I traced that through the JSTL 3.0.1 sources rather than rendering the page. Keeping the old blanking without the dead lookup:

```suggestion
			<c:set var="obsList" value="${groupMembersMap != null && groupMembersMap[obs] != null ? groupMembersMap[obs] : obs.groupMembers}" scope="request"/>
			<c:remove var="field" scope="request"/>
```

omod/src/main/java/org/openmrs/web/dwr/DWRObsService.java
```diff
@@ -317,13 +336,23 @@ public Vector<ObsListItem> getObsByPatientConceptEncounter(String personId, Stri
...
+			    Collections.singletonList(p), null, Collections.singletonList(c), null, null, null, null, null, null, null, null, true));
```
dkayiwa, line 337: Flipping `includeVoidedObs` to `true` here, and on the person-only branch at line 351, widens this past archived obs, is that part of LUI-213? The archived rows come from the helper two lines down, so the flag is not what brings them in; what it adds is every voided obs still sitting in the `obs` table. Built against your core branch, after voiding obs 7, `getObsByPatientConceptEncounter("7", null, null)` returns `... 15 9 7` on this branch against `... 15 9` on master, and encounter 3 returns `7 9` against `9`, since `getAllObs()` was non-voided while `getAllObsIncludingArchived()` adds `this.obs` unfiltered. Manage Observations has no include-voided control, so those rows simply show up, struck through.

If that is wanted, `getObsByPatientConceptEncounter_shouldIncludeArchivedObs` pins only the archived half right now; putting a voided-but-live obs in the same assertion would stop someone quietly setting the flag back to `false`. If it is not wanted, `false` here and on line 351 still leaves the archive merge working.

omod/src/main/java/org/openmrs/web/dwr/DWRObsService.java
dkayiwa: A rerun will not move this. The one that finished on Aug 5 failed with the same eight `cannot find symbol` errors, because openmrs/openmrs-core#6171 is still open and the published 3.0.0-SNAPSHOT still carries no `ObsArchiveHelper`, `Encounter.getAllObsIncludingArchived()` or `OpenmrsConstants.GP_OBS_ARCHIVE_LAST_PROCESSED_OBS_ID`.

Nothing on this side needs changing for it: I rebuilt this branch against the current head of your core branch (46e130c0) and the omod compiles with all 393 module tests passing. So it stays red until core merges and a snapshot carrying it is deployed.

omod/src/main/java/org/openmrs/web/controller/encounter/EncounterFormController.java
dkayiwa: `referenceData_shouldIncludeFullyArchivedGroupMembersInMap` does not reach the change that fixes this. I reverted only `obsDisplay.jsp` back to master on top of your head and the test still passes: it walks `groupMembersMap.entrySet()`, and that map was already correct before the fix (it is what my repro above printed as `parent=998 members=[999]`). What was broken is the `<c:when>` gate on the obs the JSP iterates.

Asserting from the `obsMap` side does cover it, since `${groupMembersMap[obs]}` is a `Map.get` on that same instance:

```java
SortedMap<FormField, List<Obs>> obsMap = (SortedMap<FormField, List<Obs>>) map.get("obsMap");
Obs parent = /* the top-level obs with obsId 998, out of obsMap.values() */;
assertFalse(parent.isObsGrouping());            // old gate: renders as a leaf, members dropped
assertNotNull(groupMembersMap.get(parent));     // new gate: recurses
```

That is worth pinning for a second reason. The `get` only resolves because `ObsArchiveHelper.convertToObs` copies the uuid and `BaseOpenmrsObject` hashes on it; the key is a different `Obs` instance from the one in `obsMap` (I printed `identityInMapKeys=false`). If core ever stops carrying the uuid across, these rows go silently missing again and nothing fails.

Line 18 of the JSP has the same hole, and it is the Toggle Voided half of the ticket. With obs 9 made a voided member of live obs 7, `obs.groupMembers` comes back `[]` while `groupMembersMap[7]` is `[9]`, so swapping the source is what makes voided members visible again, and nothing asserts that either.

omod/src/main/java/org/openmrs/web/dwr/DWRObsService.java
shivvani-r: > putting a voided-but-live obs in the same assertion

yes, this change was intentional. I edited the test to include this assertion.

omod/src/main/webapp/admin/encounters/obsDisplay.jsp
shivvani-r: will do.

dkayiwa:
@claude review

omod/src/main/java/org/openmrs/web/dwr/DWRObsService.java
```diff
@@ -356,4 +387,17 @@ public ObsListItem getObs(Integer obsId) {
 		
 		return oItem;
 	}
+
+	private void sortObservations(java.util.List<Obs> obss) {
```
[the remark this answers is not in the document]
shivvani-r: added an assertion to getObsByPatientConceptEncounter_shouldIncludeArchivedObs to guarantee that the live observation is sorted

omod/src/main/java/org/openmrs/web/dwr/DWRObsService.java
```diff
@@ -317,13 +336,25 @@ public Vector<ObsListItem> getObsByPatientConceptEncounter(String personId, Stri
...
+			    Collections.singletonList(p), null, Collections.singletonList(c), null, null, null, null, null, null, null, null, true));
+			ObsArchiveHelper archiveHelper = getObsArchiveHelperSafely();
+			if (archiveHelper != null) {
+				obss.addAll(archiveHelper.getArchivedObsByPersonIdAndConceptId(p.getPersonId(), c.getConceptId()));
+			}
+			sortObservations((java.util.List<Obs>) obss);
 		} else if (e != null) {
 			log.debug("Getting obss by encounter");
-			obss = e.getAllObs();
+			obss = e.getAllObsIncludingArchived();
```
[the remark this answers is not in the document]
shivvani-r: Thanks i made the change you suggested

omod/src/main/java/org/openmrs/web/controller/encounter/EncounterFormController.java
```diff
@@ -328,31 +343,22 @@ protected Map<String, Object> referenceData(HttpServletRequest request, Object o
...
+					obsMapToReturn.put(ff, list);
 				}
+				list.add(o);
 			}
+			map.put("groupMembersMap", groupMembersMap);
 		}
 		
 		if (log.isDebugEnabled()) {
+			log.error("Total obs: " + encounter.getAllObsIncludingArchived().size());
 			log.debug("setting obsMap in page context (size: " + obsMapToReturn.size() + ")");
```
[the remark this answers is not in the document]
shivvani-r: Thanks i deleted it

shivvani-r:
@dkayiwa I added a new commit addressing the review comments. Could you please review it again?

dkayiwa:
@claude review

dkayiwa:
@claude review

omod/src/main/java/org/openmrs/web/dwr/DWRObsService.java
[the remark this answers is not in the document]
dkayiwa: The new assertion doesn't fail when the sort goes away. I commented out all three `sortObservations` calls on 61a8ed51 and `getObsByPatientConceptEncounter_shouldIncludeArchivedObs` still passed. Unsorted order is already live-then-archived, since the live rows come back from `getObservations(...)` and the archived ones are appended after them, so `liveIndex < archivedIndex` holds either way.

What makes it bite is an archived obs that is newer than the live one, so the sorted and unsorted orders disagree. Dating 999 ahead of `liveObs` and turning the comparison round:

```java
"INSERT INTO obs_archive (obs_id, person_id, concept_id, obs_datetime, voided, uuid, creator, date_created, status) VALUES (999, 2, 21, '2027-01-01', 1, 'archive-uuid-1', 1, '2026-01-01', 'FINAL')"
...
assertTrue(archivedIndex < liveIndex, "Newer archived obs should sort ahead of the older live obs");
```

I ran that against your branch: green as it stands, and `AssertionFailedError` as soon as the `sortObservations` calls come out.

omod/src/main/java/org/openmrs/web/controller/encounter/EncounterFormController.java
dkayiwa: `referenceData_shouldIncludeFullyArchivedGroupMembersInMap` still doesn't reach either half of the JSP change. I reverted `obsDisplay.jsp` to master on top of 61a8ed51 and re-ran it, and it passes: it reads `groupMembersMap` straight off the model, and that map was already being built correctly before the gate moved.

The two cases the JSP change covers behave differently, and both are assertable from the controller's model. I checked each against your branch.

Fully archived group, which is what line 10 fixes. `convertToObs` never fills in `groupMembers`, so the parent comes back with `isObsGrouping() == false` and the old gate rendered it as a leaf:

```java
SortedMap<FormField, List<Obs>> obsMap = (SortedMap<FormField, List<Obs>>) map.get("obsMap");
Obs parent = /* the obs with obsId 998, out of obsMap.values() */;
Assertions.assertFalse(parent.isObsGrouping());
Assertions.assertNotNull(groupMembersMap.get(parent));
```

Voided member of a live parent, which is what line 18 fixes. `isObsGrouping()` is `hasGroupMembers(true)`, so it is already true here and the old gate did recurse. What it recursed over was empty. Pointing obs 9's `obs_group_id` at 7 and voiding it, then calling `referenceData` on encounter 3, gives:

```
parent.getGroupMembers()    = []         <- the old obsList source
groupMembersMap.get(parent) = [Obs #9]   <- the new one
```

shivvani-r: Thanks, I've updated the referenceData_shouldIncludeFullyArchivedGroupMembersInMap test to fetch the parent from obsMap and assert it is correctly recovered from groupMembersMap, and I've added a matching test for the 'Toggle Voided' side to cover live parents with voided children.

omod/src/main/java/org/openmrs/web/dwr/DWRObsService.java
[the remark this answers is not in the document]
shivvani-r: I've updated the test to date the archived obs ahead of the live one and reversed the assertion to guarantee the sort order is properly verified.

shivvani-r:
@dkayiwa added a new commit addressing the review comments, kindly review when you're free.

omod/src/main/java/org/openmrs/web/dwr/DWRObsService.java
dkayiwa: Core moved under this one. `cb162668` on openmrs/openmrs-core#6171 ("drop lastprocessedid Encounter archive gate") took the `obs.archive.last_processed_obs_id` check back out of `Encounter.getAllObsIncludingArchived()`, leaving just the try/catch. So the half of my earlier note that asked for that check is the part that is now wrong, and keeping it here is what makes the branches of this method disagree.

Built against that core head, with no `obs.archive.last_processed_obs_id` row present:

```
getObsByPatientConceptEncounter("2", null, null)   archived 999 returned = false
getObsByPatientConceptEncounter("2", "21", null)   archived 999 returned = false
getObsByPatientConceptEncounter(null, null, "3")   archived 999 returned = true
```

The encounter branch goes through `getAllObsIncludingArchived()`, which no longer consults the property, while these two still do. findObservation.jsp puts both searches on the same page (lines 44 and 60), so the same archived obs is visible under "by encounter" and missing under "by person".

On a started server the gate never fires anyway. `OpenmrsConstants.CORE_GLOBAL_PROPERTIES()` registers the property with a default of `"-1"` and `Context.checkCoreDataset()` writes it at startup, so the value is non-empty and the lookup always runs; I re-ran the person branch with it set to `-1` and 999 came back. What the gate does add is a way for an admin to switch archived rows off in half the UI by blanking a property whose own description invites resetting it, which is the hazard you argued against on the core PR when you kept `getObs`/`getObsByUuid` ungated by `isArchivingEnabled()`.

I would drop the property lookup and keep the rest:

```java
private ObsArchiveHelper getObsArchiveHelperSafely() {
	try {
		return Context.getRegisteredComponent("obsArchiveHelper", ObsArchiveHelper.class);
	}
	catch (APIException e) {
		// bean not registered on this core, degrade gracefully
	}
	return null;
}
```

Every query method on `ObsArchiveHelper` already catches `HibernateException` and returns an empty list, so a missing table still degrades on its own. The two `setGlobalProperty("obs.archive.last_processed_obs_id", ...)` lines in the tests go with it. The one in `EncounterFormControllerTest` is dead already: I called `referenceData` without it and the archived parent still turned up in `groupMembersMap`.

Separately, and bigger than this PR so not something to hold it for: `ObsServiceImpl.getObs` and `getObsByUuid` already fall back to the archive. The same fallback inside `getObservations(...)`, or an `includeArchived` flag on it, would let this file drop the helper lookup, the merge and `sortObservations` altogether, and would save REST and FHIR writing the same merge later. It would also get legacyui out of `org.openmrs.api.impl`, which is what let this drift happen. Worth deciding while #6171 is still open rather than after it merges.

omod/src/main/java/org/openmrs/web/dwr/DWRObsService.java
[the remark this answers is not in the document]
dkayiwa: `liveObs` still gets `new Date()` while the archive row is pinned at `2027-01-01`, so `assertTrue(archivedIndex < liveIndex)` only holds while today is earlier than that date. From 1 Jan 2027 `new Date()` is the later of the two, the sort puts the live obs first, and the test fails for everyone. That literal came out of my suggestion up-thread, so this one is on me.

I ran the same fixture with the archive dated behind `new Date()` (2025-01-01, which is the ordering that will hold from 2027 onwards) and got `archivedIndex=1 liveIndex=0`, so it goes red exactly there.

Pinning both dates keeps what the assertion is for and drops the dependency on today:

```java
liveObs.setObsDatetime(new java.text.SimpleDateFormat("yyyy-MM-dd").parse("2008-08-01"));
```

with the archive row at `'2008-09-01'` instead of `'2027-01-01'`. Against your branch that gives `archivedIndex=0 liveIndex=1`, and commenting the three `sortObservations` calls out flips it back to `archivedIndex=1 liveIndex=0`, so it still bites if the sort ever goes away.

omod/src/main/java/org/openmrs/web/dwr/DWRObsService.java
shivvani-r: Removed the code related to last_processed_obs_id, and also fixed the test like you suggested.

omod/src/main/java/org/openmrs/web/dwr/DWRObsService.java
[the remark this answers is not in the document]
shivvani-r: thanks fixed it as you suggested

shivvani-r:
@dkayiwa could you please review this again?

omod/src/test/java/org/openmrs/web/controller/encounter/EncounterFormControllerTest.java
```diff
@@ -63,4 +65,125 @@ public void onSubmit_shouldSaveANewEncounterRoleObject() throws Exception {
...
+		
+		Encounter encounter = Context.getEncounterService().getEncounter(3);
+		
+		try {
+			Context.getAdministrationService().executeSQL(
+			    "INSERT INTO obs_archive (obs_id, person_id, concept_id, encounter_id, obs_datetime, voided, uuid, creator, date_created, status) VALUES (998, 2, 21, 3, '2026-01-01', 1, 'uuid-998', 1, '2026-01-01', 'FINAL')",
+			    false);
+			Context.getAdministrationService().executeSQL(
+			    "INSERT INTO obs_archive (obs_id, person_id, concept_id, encounter_id, obs_datetime, voided, uuid, creator, date_created, status, obs_group_id) VALUES (999, 2, 21, 3, '2026-01-01', 1, 'uuid-999', 1, '2026-01-01', 'FINAL', 998)",
+			    false);
```
dkayiwa, line 81: This one needs fixing before merge, and the cause is new: core moved again today, and `1a0b5a3d9` ("Add volatile archive cache") gave `ObsArchiveHelper` a `doesArchiveHaveData()` short-circuit. Every query method on the helper returns an empty result while that flag is false, the flag lives on the singleton bean with a 60s TTL, and a raw SQL insert never flips it. So this test sees its own archive rows only when nothing before it in the same JVM has already read an empty `obs_archive`.

`referenceData_shouldIncludeVoidedGroupMembersInMap` is exactly such a read: it inserts into `obs` only, then calls `referenceData`, which caches `false`. Today JUnit happens to run this test first, so it caches `true` and the ordering never bites. Forcing the other order does. Built against the current head of your core branch (678441f4) with `MethodOrderer$Random`:

```
seed=1   voided first    -> AssertionFailedError: Archived parent should be in obsMap ==> expected: not <null>
seed=2   voided first    -> same failure
seed=3   archived first  -> green
```

A fourth test in this class, a renamed method, or a JUnit bump is enough to land on one of those orders, and the failure message points at the archive lookup rather than at the ordering, so whoever hits it starts debugging the wrong thing.

`markArchiveHasData()` is the supported way to say "a sweep just landed on this node", and core's own `ObsArchiveIntegrationTest` uses it for this. Adding it made seed=1 green for me:

```suggestion
			    false);
			Context.getRegisteredComponent("obsArchiveHelper", org.openmrs.api.impl.ObsArchiveHelper.class)
			        .markArchiveHasData();
```

`DWRObservationServiceTest.getObsByPatientConceptEncounter_shouldIncludeArchivedObs` has the same exposure and wants the same line. I put one `getObsByPatientConceptEncounter` call against an empty archive ahead of its insert, and the archived row came back missing (`rows=0`).

omod/src/main/java/org/openmrs/web/dwr/DWRObsService.java
```diff
@@ -28,7 +32,10 @@
 import org.openmrs.Obs;
 import org.openmrs.Person;
 import org.openmrs.api.AdministrationService;
+import org.openmrs.api.APIException;
 import org.openmrs.api.context.Context;
+import org.openmrs.api.impl.ObsArchiveHelper;
+import org.openmrs.util.OpenmrsConstants;
 import org.openmrs.util.OpenmrsUtil;
```
dkayiwa: Leftover from dropping the `last_processed_obs_id` gate: `OpenmrsConstants` was the only reason for this import, and nothing in the file mentions it any more (`OpenmrsUtil` on the next line is still used, by `sortObservations`). It is not on master either, so it arrived with this PR.

```suggestion
import org.openmrs.util.OpenmrsUtil;
```

omod/src/main/java/org/openmrs/web/dwr/DWRObsService.java
```diff
@@ -317,13 +334,26 @@ public Vector<ObsListItem> getObsByPatientConceptEncounter(String personId, Stri
...
 		if (p != null && c != null) {
 			log.debug("Getting obss with patient and concept");
-			obss = Context.getObsService().getObservationsByPersonAndConcept(p, c);
+			obss = new ArrayList<Obs>(Context.getObsService().getObservations(
+			    Collections.singletonList(p), null, Collections.singletonList(c), null, null, null, null, null, null, null, null, true));
+			ObsArchiveHelper archiveHelper = getObsArchiveHelperSafely();
+			if (archiveHelper != null) {
+				obss.addAll(archiveHelper.getArchivedObsByPersonIdAndConceptId(p.getPersonId(), c.getConceptId()));
+			}
+			sortObservations((List<Obs>) obss);
```
dkayiwa: Worth declaring `obss` as `List<Obs>` at line 333 instead, since all three branches now assign an `ArrayList` and these three casts stop being needed. As it stands the cast is unchecked, so a later branch that assigns the `Set` straight from `getAllObsIncludingArchived()`, rather than wrapping it the way line 346 does, would compile fine against `Collection<Obs>` and then throw `ClassCastException` here at runtime.

shivvani-r: changed it!

omod/src/main/java/org/openmrs/web/dwr/DWRObsService.java
shivvani-r: removed it thanks

omod/src/test/java/org/openmrs/web/controller/encounter/EncounterFormControllerTest.java
shivvani-r: I've added markArchiveHasData() here, and applied the same fix over in DWRObservationServiceTest as well.

shivvani-r:
@dkayiwa could you please review this?

omod/src/main/java/org/openmrs/web/dwr/DWRObsService.java
dkayiwa: Coming back to the second half of that note, because two more pages in this module are still blind to the archive, and both exist specifically to show voided obs.

`admin/observations/personObs.form` ("all observations for a person") renders voided rows with a strikethrough and a void-info column, and it is reachable from the Graphs tab on the patient dashboard, where `ObsTableWidget` links each concept header at `personObs.form?personId=..&conceptId=..`. `PersonObsFormController.formBackingObject` builds its list with the same call this PR wraps twice:

```java
os.getObservations(Collections.singletonList(person), null, concepts, null, null, null, null, null, null, null, null, true);
```

`QuickReportServlet.doVoidedObs`, behind the "Voided Obs" option on `admin/maintenance/quickReport.form`, does the same with a date range and then keeps only the voided rows.

Built against the current head of your core branch (659f7e5d) with obs 994 archived for person 2:

```
manage observations (DWR)  ->  994 returned
personObs.form             ->  0 rows, 994 missing
quickReport VOIDED OBS     ->  994 missing
```

So once this lands the archived row is visible on the two pages LUI-213 names and still missing from the two whose whole job is voided obs. Writing the merge out per call site is what leaves those behind, which is the argument for the fallback sitting in `getObservations`.

For this PR I would note it on the ticket rather than grow the diff, since "they completely disappear from the Legacy UI" reads as covered otherwise. If you would rather close personObs.form here, `getArchivedObsByPersonId` gives it to you in the same three lines as the person branch above.

omod/src/main/java/org/openmrs/web/dwr/DWRObsService.java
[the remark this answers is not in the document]
dkayiwa: The pinning still reaches only one of the three branches. `getObsByPatientConceptEncounter_shouldIncludeArchivedObs` calls `("2", "21", null)`, so `getArchivedObsByPersonId` on the person-only branch and `getAllObsIncludingArchived` on the encounter branch have nothing on them, and the encounter branch's `sortObservations` is the one that went in last.

Both do work. Against the current core head with obs 994 archived for person 2 on encounter 3:

```
("2","21",null)  ->  994v
("2",null,null)  ->  994v
(null,null,"3")  ->  994v 9 7
```

Commenting out the encounter branch's `sortObservations` turns that last line into `7 9 994v`, so an index assertion there bites the same way the one you already have does. Two more calls in this test would cover both branches.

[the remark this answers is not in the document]
shivvani-r: Added tests to cover all three branches.

omod/src/main/java/org/openmrs/web/dwr/DWRObsService.java
shivvani-r: I've updated PersonObsFormController  to include archived observations similar to the person branch in DWRObsService.

shivvani-r:
@dkayiwa could you review this?

omod/src/main/java/org/openmrs/web/controller/observation/PersonObsFormController.java
```diff
@@ -52,8 +54,21 @@ protected CommandObject formBackingObject(HttpServletRequest request) throws Exc
...
+			ObsArchiveHelper archiveHelper = Context.getRegisteredComponent("obsArchiveHelper", ObsArchiveHelper.class);
+			if (concept != null) {
+				ret.addAll(archiveHelper.getArchivedObsByPersonIdAndConceptId(person.getPersonId(), concept.getConceptId()));
+			} else {
+				ret.addAll(archiveHelper.getArchivedObsByPersonId(person.getPersonId()));
+			}
+		}
+		catch (APIException e) {
+			// bean not registered on this core, degrade gracefully
+		}
```
dkayiwa, line 70: This needs a null guard before it goes in. `person` here is whatever `getPerson(Integer.valueOf(...))` returned, so a well-formed personId that no longer resolves comes back null, and `person.getPersonId()` throws on it.

Built against the current head of your core branch, hitting `personObs.form?personId=999999`:

```
master   ->  renders an empty table, 0 rows
this PR  ->  java.lang.NullPointerException: Cannot invoke "org.openmrs.Person.getPersonId()" because "person" is null
```

The call this replaced tolerated the null (`getObservations` with a single-null list just came back empty), so a stale bookmark, or a person who has since been purged, goes from a blank page to a 500. `catch (APIException e)` does not catch it either. Both person branches in `DWRObsService` do carry that guard, `if (p != null && c != null)` and `else if (p != null)`, which is why they are fine and this one is not.

```suggestion
		if (person != null) {
			try {
				ObsArchiveHelper archiveHelper = Context.getRegisteredComponent("obsArchiveHelper", ObsArchiveHelper.class);
				if (concept != null) {
					ret.addAll(archiveHelper.getArchivedObsByPersonIdAndConceptId(person.getPersonId(), concept.getConceptId()));
				} else {
					ret.addAll(archiveHelper.getArchivedObsByPersonId(person.getPersonId()));
				}
			}
			catch (APIException e) {
				// bean not registered on this core, degrade gracefully
			}
		}
```

omod/src/main/java/org/openmrs/web/controller/observation/PersonObsFormController.java
```diff
@@ -52,8 +54,21 @@ protected CommandObject formBackingObject(HttpServletRequest request) throws Exc
 		}
 		
 		ObsService os = Context.getObsService();
-		List<Obs> ret = os.getObservations(Collections.singletonList(person), null, concepts, null, null, null, null, null,
-		    null, null, null, true);
+		List<Obs> ret = new java.util.ArrayList<Obs>(os.getObservations(Collections.singletonList(person), null, concepts, null, null, null, null, null,
```
dkayiwa, line 57: Worth a test before this merges. `personObs.form` is the only one of the data-fetching changes here that nothing pins, so the merge just below can be dropped in a later refactor with the whole suite still green, and the page quietly goes back to hiding archived obs.

A controller test in the shape of `DWRObservationServiceTest` reaches it. Seeding obs 994 archived for person 2 and calling `formBackingObject` against the current core head:

```
personId=2               ->  994v
personId=2&conceptId=21  ->  994v
```

so one test class covers both sides of the `concept != null` split. It wants the `markArchiveHasData()` line for the same reason `EncounterFormControllerTest` does: without it, whether the archive is queried at all depends on whether something earlier in the same JVM already cached an empty `obs_archive`.

omod/src/main/java/org/openmrs/web/dwr/DWRObsService.java
dkayiwa: `QuickReportServlet.doVoidedObs` is still on the plain `getObservations(...)` call, and LUI-213 has no note about it either, so the ticket still reads as fully covered.

Against the current core head, with obs 994 archived for person 2 and a live voided obs in the same date range as a control:

```
quickReport VOIDED OBS  ->  78 (the live voided one)
                            994 missing
```

That leaves the "Voided Obs" option on `admin/maintenance/quickReport.form` as the one page whose whole job is showing voided obs and which still cannot see them. Either the same three lines you just used in `PersonObsFormController`, or a line on the ticket putting it out of scope, closes this out.

omod/src/main/java/org/openmrs/web/controller/observation/PersonObsFormController.java
shivvani-r: Wrapped the archive lookup in if (person != null) as suggested.

omod/src/main/java/org/openmrs/web/controller/observation/PersonObsFormController.java
shivvani-r: Added PersonObsFormControllerTest covering both the person-only and person+concept branches, with markArchiveHasData().

omod/src/main/java/org/openmrs/web/dwr/DWRObsService.java
dkayiwa: Still open at 9f6f8ccf: `QuickReportServlet` is unchanged from master, and LUI-213 has picked up no scope note, so "Voided Obs" on `admin/maintenance/quickReport.form` is still the one page whose whole job is voided obs and which cannot see them.

The blindness is in the call rather than in the servlet, so it will not come right on its own once core lands. Built against the current head of your core branch (171a1a5b), with obs 992 voided in `obs` as a control and obs 993 in `obs_archive`, both dated inside the range:

```
os.getObservations(null, ..., 2008-01-01, 2009-01-01, true)  ->  rows=11, 992 returned, 993 missing
```

That is line 304's call verbatim. `ObsServiceImpl.getObservations` hands straight to `dao.getObservations`, and the archive fallback core added lives in `getObs` and `getObsByUuid` only.

Either exit from before still closes it: the three lines you used in `PersonObsFormController`, or a line on the ticket.

PR #271, opened 2026-09-08
(docs) Replace the dead Travis badge with the GitHub Actions one

denniskigen:
The README badge still points at travis-ci.org, which shut down and now redirects to a 404, so the badge renders blank. This module builds on GitHub Actions via build.yml, so this points the badge and its link there.

Heads up that the badge will show red right now: the latest build.yml run on master failed on 2026-09-03. In this repo it's the Java 25 build job itself, not just the OWASP dependency check.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,5 @@
 # openmrs-module-legacyui
-[![Build Status](https://github.com/openmrs/openmrs-module-legacyui/actions/workflows/build.yml/badge.svg)](https://github.com/openmrs/openmrs-module-legacyui/actions/workflows/build.yml) [![Coverage Status](https://coveralls.io/repos/github/openmrs/openmrs-module-legacyui/badge.svg?branch=master)](https://coveralls.io/github/openmrs/openmrs-module-legacyui?branch=master)
+[![Build Status](https://github.com/openmrs/openmrs-module-legacyui/actions/workflows/build.yml/badge.svg)](https://github.com/openmrs/openmrs-module-legacyui/actions/workflows/build.yml)
 
 - The legacy user interface for OpenMRS Platform 2.x is chiefly comprised of administrative functions and the patient dashboard. 
 - A new and more contemporary UI has been introduced via a UI framework and the legacy UI is kept around for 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,5 +1,5 @@
 # openmrs-module-legacyui
-[![Build Status](https://travis-ci.org/openmrs/openmrs-module-legacyui.svg?branch=master)](https://travis-ci.org/openmrs/openmrs-module-legacyui) [![Coverage Status](https://coveralls.io/repos/github/openmrs/openmrs-module-legacyui/badge.svg?branch=master)](https://coveralls.io/github/openmrs/openmrs-module-legacyui?branch=master)
+[![Build Status](https://github.com/openmrs/openmrs-module-legacyui/actions/workflows/build.yml/badge.svg)](https://github.com/openmrs/openmrs-module-legacyui/actions/workflows/build.yml) [![Coverage Status](https://coveralls.io/repos/github/openmrs/openmrs-module-legacyui/badge.svg?branch=master)](https://coveralls.io/github/openmrs/openmrs-module-legacyui?branch=master)
```
dkayiwa: While you're on this line, the Coveralls badge sitting next to it has the same problem, just less visibly. It still renders, so nothing looks broken, but the coveralls.io API reports `calculated_at: 2018-03-17` against commit `afd8278`, and nothing has fed it since Travis went away. So the README is advertising 25% coverage measured in March 2018, and unlike the Travis badge there is no visual tell that the number is stale.

I'd drop it here rather than repoint it, because there is no live coverage source to point at yet. Codecov reads `unknown` for this repo (the `Upload Coverage` workflow fires on every build but skips its upload step for want of a `coverage-reports` artifact, on green master builds too), and SonarCloud has no `coverage` measure for the project either. Once JaCoCo's `report` goal is actually wired in and the build stages the report, a Codecov badge like the one in openmrs-core's README would be the thing to add.

Entirely optional as far as this PR is concerned, and fine as a follow-up if you'd rather keep it to the one badge.

```suggestion
[![Build Status](https://github.com/openmrs/openmrs-module-legacyui/actions/workflows/build.yml/badge.svg)](https://github.com/openmrs/openmrs-module-legacyui/actions/workflows/build.yml)
```

denniskigen: Good catch. Dropped it here. Agreed the Codecov badge can come once JaCoCo's report goal is actually wired in.

